PR #20, opened 2020-04-22
Return empty NotFound when Details/Failure is nil

shawnhathaway:
Fix for when details are not specified - See this test in SDK https://github.com/temporalio/temporal-go-sdk/blob/v0.22.0/internal/grpc_dialer_test.go#L39

### DIFF
--- a/serviceerror/convert.go
+++ b/serviceerror/convert.go
@@ -91,11 +91,13 @@ func FromStatus(st *status.Status) error {
 	f := extractFailure(st)
 	switch st.Code() {
 	case codes.NotFound:
-		if f != nil {
-			return newNotFound(st, f.(*failure.NotFound))
+		if f == nil{
+			return newNotFound(st, nil)
 		}
-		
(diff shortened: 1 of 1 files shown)


serviceerror/convert.go
```diff
@@ -91,7 +91,11 @@ func FromStatus(st *status.Status) error {
 	f := extractFailure(st)
 	switch st.Code() {
 	case codes.NotFound:
-		return newNotFound(st, f.(*failure.NotFound))
+		if f != nil {
+			return newNotFound(st, f.(*failure.NotFound))
+		}
+		
+		return newNotFound(st, &failure.NotFound{})
```
alexshtin, line 98:
```suggestion
		if f == nil{
			return newNotFound(st, nil)
		}
		switch f := f.(type) {
		case *failure.NotFound:
			return newNotFound(st, f)
		}
```

alexshtin: Just to match all the rest in this file.

PR #64, opened 2020-07-31
CI/CD: updating with the most recent API proto updates

markmark206:


markmark206:
(this change was generated by tinycicd pipeline, ran this by alex, merging).

PR #83, opened 2021-05-26
Add ServerConviguration service error

alexshtin:
Add `ServerConviguration` service error.

### DIFF
--- a/serviceerror/serverConfiguration.go
+++ b/serviceerror/serverConfiguration.go
@@ -1,0 +1,68 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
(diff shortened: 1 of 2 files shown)


serviceerror/serverConfiguration.go
```diff
@@ -0,0 +1,68 @@
...
+func (e *ServerConfiguration) Error() string {
+	return e.Message
+}
+
+func (e *ServerConfiguration) Status() *status.Status {
+	if e.st != nil {
+		return e.st
+	}
+
+	st := status.New(codes.FailedPrecondition, e.Message)
```
alexshtin, line 56: Rationale behind `FailedPrecondition` is in gRPC doc:
```
 	// FailedPrecondition indicates operation was rejected because the
	// system is not in a state required for the operation's execution.
	// For example, directory to be deleted may be non-empty, an rmdir
	// operation is applied to a non-directory, etc.
	//
	// A litmus test that may help a service implementor in deciding
	// between FailedPrecondition, Aborted, and Unavailable:
	//  (a) Use Unavailable if the client can retry just the failing call.
	//  (b) Use Aborted if the client should retry at a higher-level
	//      (e.g., restarting a read-modify-write sequence).
	//  (c) Use FailedPrecondition if the client should not retry until
	//      the system state has been explicitly fixed. E.g., if an "rmdir"
	//      fails because the directory is non-empty, FailedPrecondition
	//      should be returned since the client should not retry unless
	//      they have first fixed up the directory by deleting files from it.
	//  (d) Use FailedPrecondition if the client performs conditional
	//      REST Get/Update/Delete on a resource and the resource on the
	//      server does not match the condition. E.g., conflicting
	//      read-modify-write on the same resource.
```

alexshtin:
"We will take another route" (c) Lenin.

PR #85, opened 2021-06-22
Use & pin go mock to 1.5.0

wxing1292:
**What changed?**
* Use & pin go mock to 1.5.0

**Why?**
go mock 1.6.0 will break server

**How did you test it?**
N/A

**Potential risks**
N/A


Makefile
```diff
@@ -89,6 +89,7 @@ copyright:
 update-dependencies:
 	printf $(COLOR) "Update go dependencies..."
 	go get -u -t ./...
+	go get github.com/golang/mock/mockgen@v1.5.0
```
alexshtin, line 92:
```suggestion
	go get -u -t github.com/golang/mock/mockgen@v1.5.0 ./...
```

PR #87, opened 2022-01-25
Add PendingWorkflowTaskInfo in DescribeWorkflowExecutionResponse

yycptt:
**What changed?**
- Add PendingWorkflowTaskInfo in DescribeWorkflowExecutionResponse
- For https://github.com/temporalio/api/pull/139
- Will update submodule after the above pr is landed.

**Why?**
- Increase visibility into workflow task retrying. Current when workflow task is keep retrying, there's no progress in workflow history. There's also no info about it when describing the workflow. It's hard for user to know if workflow is making progress or not (workflow task is retrying or not)

**How did you test it?**

**Potential risks**


cretz:
I don't think we do manual PRs here, but rather once your PR to https://github.com/temporalio/api is merged, it'll automatically be generated then merged here.

yycptt:
> I don't think we do manual PRs here, but rather once your PR to https://github.com/temporalio/api is merged, it'll automatically be generated then merged here.

Got it! Good to know, thanks!

PR #90, opened 2022-03-20
Add AlreadyExists Failure

Spikhalskiy:
**What changed?**
Add a basic AlreadyExists Service Error

**Why?**
I want to return a descriptive code from the operator API in case the entity attempted to register already exists. Currently, we use INVALID_ARGUMENT and it's not technically correct and hard to be parsed by the clients.

See https://github.com/temporalio/api/pull/161

### DIFF
--- a/serviceerror/convert.go
+++ b/serviceerror/convert.go
@@ -78,6 +78,7 @@ func FromStatus(st *status.Status) error {
 	case codes.Unknown:
 		// Unwrap error message from unknown error.
 		return errors.New(st.Message())
+
 	// Unsupported codes.
 	case codes.OutOfRange,
 		codes.Unauthenticated:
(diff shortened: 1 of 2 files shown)


serviceerror/convert.go
```diff
@@ -75,9 +75,12 @@ func FromStatus(st *status.Status) error {
 		return newUnavailable(st)
 	case codes.Unimplemented:
 		return newUnimplemented(st)
+	case codes.AlreadyExists:
```
cretz, line 78: This breaks later code that uses already exists below for specific namespace, workflow execution, and cancellation errors. If we must have a generic already-exists, add as an else to those later.

Spikhalskiy: Good catch, thanks!

cretz (approved):
:+1: Would like approval from one member of server team

PR #94, opened 2022-05-13
Don't return InvalidArgument if error details are of unknown type

alexshtin:
**What changed?**
Don't return `InvalidArgument` if error details are of unknown type.

**Why?**
Previously if new error details type is added then `InvalidArgument` was returned no matter of actual status code. This is because `extractErrorDetails` returned and error and this [error](https://github.com/temporalio/gogo-protobuf/blob/master/types/any.go#L96) was treated as fatal. This made adding new error details very inconvenient. With thss PR, in case of unknown error details type (and any other deserialization error), `serviceerror` will be built based on status code only which is more reasonable.

**How did you test it?**
New unit tests.

**Potential risks**
Some weird deserialization errors will be swallowed.

### DIFF
--- a/serviceerror/convert.go
+++ b/serviceerror/convert.go
@@ -63,7 +63,7 @@ func FromStatus(st *status.Status) error {
 		return nil
 	}
 
-	// Simple case. Code to serviceerror is one to one mapping and there are no error details.
+	// Simple case. `st.Code()` to `serviceerror` is one to one mapping and there are no error details.
 	switch st.Code() {
 	case codes.DataLoss:
 		return newDataLoss(st)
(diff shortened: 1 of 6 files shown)


serviceerror/convert.go
```diff
@@ -87,58 +87,52 @@ func FromStatus(st *status.Status) error {
 	}
 
 	errDetails := extractErrorDetails(st)
-
-	// If there was an error during details extraction, it will go to errDetails.
-	if err, ok := errDetails.(error); ok {
```
cretz, line 92: Lets add a unit test for this. Here's how I made an error of unknown type:

```go
err := status.ErrorProto(&statuspb.Status{
	Code:    int32(codes.NotFound),
	Message: "some error!",
	Details: []*anypb.Any{{TypeUrl: "type.googleapis.com/some.unknown.Type"}},
})
```

alexshtin: Added.

cretz (approved):

PR #100, opened 2022-12-02
Initial version.

robholland:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**


cmd/generateproxy/interceptor.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+
+	// Return whether this Type can (eventually) lead to Payload(s)
+	// This is used in the walk logic above so that our encoding handles intermediate Types between our Request/Response objects and Payload(s)
+	return record.Matches
+}
+
+func generateInterceptor(cfg config) {
+	conf := &packages.Config{Mode: packages.NeedImports | packages.NeedTypes | packages.NeedTypesInfo}
+	pkgs, err := packages.Load(conf, "go.temporal.io/api/workflowservice/v1")
```
cretz, line 314: As discussed in Slack, need to have a `replace` in go.mod for this to work with latest local changes probably

proxy/interceptor.go
```diff
@@ -0,0 +1,1632 @@
...
+	interactionpb "go.temporal.io/api/interaction/v1"
+	querypb "go.temporal.io/api/query/v1"
+	schedulepb "go.temporal.io/api/schedule/v1"
+	workflowpb "go.temporal.io/api/workflow/v1"
+	workflowservicepb "go.temporal.io/api/workflowservice/v1"
+)
+
+type VisitPayloadsContext struct {
+	context.Context
+	Parent interface{}
```
cretz, line 49: Should change to `proto.Message` IMO and it should be set to the `obj` value before each non-payload recursive call if that obj is a proto message. If this is too annoying though, we can remove this for now.

proxy/interceptor.go
```diff
@@ -0,0 +1,1632 @@
...
+}
+
+type VisitPayloadsOptions struct {
+	// Context is the same for every call of a visit, callers should not store it. This must never
+	// return an empty set of payloads.
+	Visitor              func(*VisitPayloadsContext, []*commonpb.Payload) ([]*commonpb.Payload, error)
+	SkipSearchAttributes bool
+}
+
+func VisitPayloads(ctx context.Context, msg interface{}, options VisitPayloadsOptions) error {
```
cretz, line 61:
```suggestion
func VisitPayloads(ctx context.Context, msg proto.Message, options VisitPayloadsOptions) error {
```

proxy/interceptor.go
```diff
@@ -0,0 +1,1632 @@
...
+
+		}
+	}
+
+	return nil
+}
+
+func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grpc.UnaryClientInterceptor, error) {
+	return func(ctx context.Context, method string, req, response interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		err := VisitPayloads(ctx, req, *options.Outbound)
```
cretz, line 1620: Let's skip this visit if req is not proto.Message or if options.Outbound is nil

proxy/interceptor.go
```diff
@@ -0,0 +1,1632 @@
...
+		if err != nil {
+			return err
+		}
+
+		err = invoker(ctx, method, req, response, cc, opts...)
+		if err != nil {
+			return err
+		}
+
+		return VisitPayloads(ctx, response, *options.Inbound)
```
cretz, line 1630: Let's skip this visit if response is not proto.Message or if options.Inbound is nil

cmd/generateproxy/interceptor.go
```diff
@@ -0,0 +1,374 @@
...
+		}
+	}
+
+	buf := &bytes.Buffer{}
+
+	err = interceptorTemplate.Execute(buf, pruneRecords(records))
+	if err != nil {
+		log.Fatal(err)
+	}
+	src, err := imports.Process(interceptorFile, buf.Bytes(), nil)
```
cretz, line 352: Lets make sure we do a `format.Source` for this too

cretz (commented):
This looks great! Lets get a test or two in here and if you don't mind a GH action to run on each PR just in case (w/ test and lint). I can help here if needed.

cmd/generateproxy/interceptor.go
robholland: `imports.Process` also formats.

robholland: Oh. Hmm, maybe I misunderstood. I think it only formats the imports. Will put `format.Source` back in also.

robholland:
Yep I'll copy/amend the SDK unit tests over this afternoon.

proxy/interceptor.go
```diff
@@ -0,0 +1,1739 @@
...
+	"go.temporal.io/api/history/v1"
+	"go.temporal.io/api/interaction/v1"
+	"go.temporal.io/api/query/v1"
+	"go.temporal.io/api/schedule/v1"
+	"go.temporal.io/api/workflow/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	"google.golang.org/grpc"
+)
+
+type VisitPayloadsContext struct {
```
cretz, line 47: Let's add godoc on all exported types and funcs

cretz (approved):

cretz:
@robholland - we have another use case from https://github.com/temporalio/sdk-go/issues/930. They want to visit every `temporal.api.failure.v1.Failure` in the proxy. I know it's kinda annoying to ask at last minute, but think we can genericize this a bit? I am wondering if now is the time to do this.

Maybe change `VisitPayloads` to `VisitProtos` (I understand we limit to only where payloads are today, is it that troubling to change to walk all protos?). We could still have separate "payload visitor" and "failure visitor" though having some kind of common proto visitor makes more sense, even if it's just `interface{}` because sometimes it's a slice instead of a `proto.Message`.

Any design ideas on how best to incorporate failure visiting into the proxy? I have some more detailed thoughts.

bergundy:
> @robholland - we have another use case from https://github.com/temporalio/sdk-go/issues/930. They want to visit every temporal.api.failure.v1.Failure in the proxy. I know it's kinda annoying to ask at last minute, but think we can genericize this a bit? I am wondering if now is the time to do this.

Let’s do this please while we’re touching all of this and avoid the risk of more compilation breaking changes for the Go SDK.

cmd/generateproxy/interceptor.go
```diff
@@ -0,0 +1,491 @@
...
+
+	return nil
+}
+
+func visitFailures(ctx *VisitFailuresContext, options *VisitFailuresOptions, objs ...interface{}) error {
+	for _, obj := range objs {
+		switch o := obj.(type) {
+			case *failure.Failure:
+				if o == nil { continue }
+				if err := options.Visitor(ctx, o); err != nil { return err }
```
cretz, line 228: You need to visit the cause too if present _after_ visiting this

proxy/interceptor.go
```diff
@@ -0,0 +1,2285 @@
...
+type VisitFailuresContext struct {
+	context.Context
+	// The parent message for this failure.
+	Parent proto.Message
+}
+
+// VisitFailuresOptions configure visitor behaviour.
+type VisitFailuresOptions struct {
+	// Context is the same for every call of a visit, callers should not store it. This must never
+	// return an empty set of payloads.
```
cretz, line 113:
```suggestion
	// Context is the same for every call of a visit, callers should not store it.
```

cretz (approved):
LGTM, but would like @alexshtin and/or @yiminc to take a look

cretz:
@robholland - we just noticed a bug in the Go SDK version of this where the `map[string]*commonpb.Payloads` (note the plural) isn't in the switch statement. Can you add it here?

robholland:
Added.

cmd/generateproxy/service.go
```diff
@@ -0,0 +1,133 @@
...
+type WorkflowServiceProxyOptions struct {
+	Client workflowservice.WorkflowServiceClient
+}
+
+type workflowServiceProxyServer struct {
+	workflowservice.UnimplementedWorkflowServiceServer
+	client workflowservice.WorkflowServiceClient
+}
+
+// NewWorkflowServiceProxyServer creates a WorkflowServiceServer suitable for registering with a GRPC Server. Requests will
```
alexshtin, line 62:
```suggestion
// NewWorkflowServiceProxyServer creates a WorkflowServiceServer suitable for registering with a gRPC Server. Requests will
```

please, everywhere...

cmd/generateproxy/interceptor.go
```diff
@@ -0,0 +1,493 @@
...
+
+const InterceptorTemplateText = Header + `
+package proxy
+
+import (
+	"context"
+	"fmt"
+
+	"google.golang.org/grpc"
+	proto "github.com/gogo/protobuf/proto"
```
alexshtin, line 49:
```suggestion
	"github.com/gogo/protobuf/proto"
```

cmd/generateproxy/service.go
```diff
@@ -0,0 +1,133 @@
...
+	service := scope.Lookup("UnimplementedWorkflowServiceServer")
+	if _, ok := service.(*types.TypeName); ok {
+		for _, meth := range typeutil.IntuitiveMethodSet(service.Type(), nil) {
+			if !meth.Obj().Exported() {
+				continue
+			}
+
+			name := meth.Obj().Name()
+			sig := meth.Obj().Type().(*types.Signature)
+			fmt.Fprintf(buf, "\nfunc (wh *workflowServiceProxyServer) %s %s %s {\n", name, types.TypeString(sig.Params(), qual), types.TypeString(sig.Results(), qual))
```
alexshtin, line 104:
```suggestion
			fmt.Fprintf(buf, "\nfunc (s *workflowServiceProxyServer) %s %s %s {\n", name, types.TypeString(sig.Params(), qual), types.TypeString(sig.Results(), qual))
```

alexshtin (commented):
Left some nit style comments. Didn't go much in generator code, but generated code looks good. No human should review reflection and reflection generated code though.

alexshtin (approved):

PR #101, opened 2023-01-06
Add gogoslick mapping for Google's any.proto

mmcshane:
**What changed?**
Add gogoslick mapping for Google's any.proto

**Why?**
Required to use the Any type. Use of Google version of this type fails to compile because gogo expects certain member functions.

**How did you test it?**
Compile upstream api that has been modified to include an Any field.

**Potential risks**
Low - this mapping is currently unused but will be used shortly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ go-grpc: clean $(PROTO_OUT)
 
 gogo-grpc: clean $(PROTO_OUT)
 	printf $(COLOR) "Compiling for gogo-gRPC..."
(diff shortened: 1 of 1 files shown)


cretz:
Ug, I wish we could leave this library and use https://pkg.go.dev/google.golang.org/protobuf/types/known/anypb like the rest of the world.

cretz (approved):
Approved begrudgingly.

alexshtin (approved):

PR #105, opened 2023-01-24
Downgrade grpc and genproto

bergundy:
Downgrading grpc seems to fix `make check`, didn't look into the root of the issue.


bergundy:
Closing in favor of #104

PR #106, opened 2023-02-02
Update go.sum before building proxygen

mmcshane:
**What changed?**
Run `go mod tidy` before running the proxy codegen

**Why?**
The unpinned versions (e.g. we depend on the latest version of google.golang.org/genproto) change frequently and if the go.sum isn't updated, builds will fail.

**How did you test it?**
Running locally

**Potential risks**


mmcshane:
FFS

>  go mod tidy: go.mod file indicates go 1.18, but maximum supported version is 1.17

cretz (approved):
Thanks!

PR #112, opened 2023-06-26
HTTP API helpers

cretz:
**What changed?**

Implements helpers for HTTP API as proposed at https://github.com/temporalio/proposals/pull/79, primarily around payload formatting. Specifically:

* Added `go.temporal.io/api/internal/temporaljsonpb` package
  * Added `JSONPBMaybeMarshaler` and `JSONPBMaybeUnmarshaler` interfaces which protos can implement for custom, optional JSON support
  * A copy of https://github.com/gogo/protobuf/blob/v1.3.2/jsonpb/jsonpb.go but with support for those interfaces
* Added `go.temporal.io/api/internal/temporalgateway` package
  * Implements gogo-supported equivalent of https://pkg.go.dev/github.com/grpc-ecosystem/grpc-gateway/runtime#JSONPb akin to https://github.com/gogo/gateway but with `JSONPBMaybeMarshaler`/`JSONPBMaybeUnmarshaler` support
* Added `go.temporal.io/api/internal/temporalcommonv1` package
  * Implements `JSONPBMaybeMarshaler` and `JSONPBMaybeUnmarshaler` for `Payloads` and `Payload` to support shorthand JSON representation
  * `Makefile` updated to copy code written here to `go.temporal.io/api/common/v1` where it lives alongside generated code (instead of trying to write code directly alongside generated code)
* Updated `go.temporal.io/api/proxy`
  * Added `JSONPBMarshaler`/`NewJSONPBMarshaler` and `JSONPBUnmarshaler`/`NewJSONPBUnmarshaler` to create marshalers to support maybe-JSON interfaces (so for Payload shorthand)
  * Added `NewGRPCGatewayJSONPBMarshaler` to create a https://pkg.go.dev/github.com/grpc-ecosystem/grpc-gateway/runtime#Marshaler from the above marshaler/unmarshaler
* Added tests for shorthand payloads and gRPC gateway
* Updated `Makefile` to support building with gRPC gateway + Google HTTP annotations

This is ready for review. This will not have the generated code updates until the API PR is merged. Related PRs:

* Depends on https://github.com/temporalio/api/pull/301
* Depended on by https://github.com/temporalio/temporal/pull/4543


proxy/gateway_test.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	workflowservice.RegisterWorkflowServiceServer(t.grpcServer, t)
+	go func() { t.serveErrCh <- t.grpcServer.Serve(t.grpcListener) }()
+
+	// Start HTTP server
+	if t.httpListener, err = net.Listen("tcp", "127.0.0.1:0"); err != nil {
+		t.grpcServer.Stop()
+		return nil, err
+	}
+	if err = workflowservice.RegisterWorkflowServiceHandlerServer(ctx, serveMux, t); err != nil {
```
cretz, line 205: I will push the generated gateway code when https://github.com/temporalio/api/pull/301 is reviewed

proxy/marshal_test.go
```diff
@@ -0,0 +1,191 @@
...
+	assertPayloadData(t, memo.Fields["some-string"], "json/plain", `"string"`)
+	assertPayloadData(t, memo.Fields["some-float"], "json/plain", "1.23")
+	assertPayloadData(t, memo.Fields["some-null"], "binary/null", "")
+	assertPayloadData(t, memo.Fields["some-bytes"], "binary/plain", "bytes")
+	assertPayloadJSONData(t, memo.Fields["some-proto"], "json/protobuf", `{"name":"workflow-name"}`)
+	require.Equal(t, "temporal.api.common.v1.WorkflowType", string(memo.Fields["some-proto"].Metadata["messageType"]))
+	assertPayloadData(t, memo.Fields["some-raw-payload"], "my-encoding", "raw-value")
+}
+
+func TestShorthandPayloadAmbiguity(t *testing.T) {
```
cretz, line 77: @josh-berry - you might find this test valuable to read

josh-berry: Got it, makes sense, thanks!

MichaelSnowden (commented):
If this depends on open PRs, please make a stacked PR to make review easier (e.g. push all the dependencies to some arbitrary branch and then set that as the base here).

cretz:
> If this depends on open PRs, please make a stacked PR to make review easier (e.g. push all the dependencies to some arbitrary branch and then set that as the base here).

It's more than that. It depends on a submodule from another repo being updated and a Makefile rerun (or rather a GitHub workflow triggered manually probably). You cannot just merge the other and then this magically is ok to merge. The PRs are in different repos and are inherently different things to review. Not sure how these different-repo PRs can be made any easier to review.

internal/temporalcommonv1/message.go
```diff
@@ -0,0 +1,38 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package common
+
+// These are dummy messages, never used
+
+type Payloads struct {
```
Quinn-With-Two-Ns, line 27: Dumb question, why do we need these?

Quinn-With-Two-Ns: Oh I guess because the "real" types on exist in the destination folder your copying into

cretz: Exactly. I struggled with "write code over there and make the special file excluded from the wipe-and-regen process", or "write over here and copy". I chose the latter. There are pros/cons to both approaches.

Quinn-With-Two-Ns (approved):
I'm good with the approach taken to write over and copy. It's hard to verify all the marshalling code is 100% correct, but I don't see anything obviously wrong

feedmeapples (approved):

alexshtin (approved):

internal/temporalgateway/marshal_jsonpb.go
```diff
@@ -0,0 +1,281 @@
...
+	"reflect"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	jsonpb "go.temporal.io/api/internal/temporaljsonpb"
+)
+
+// JSONPb is a Marshaler which marshals/unmarshals into/from JSON
+// with the "github.com/golang/protobuf/jsonpb".
+// It supports fully functionality of protobuf unlike JSONBuiltin.
```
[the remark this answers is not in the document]
cretz: From the top of this file:

```go
// This file taken from
// https://github.com/grpc-ecosystem/grpc-gateway/blob/v1.16.0/runtime/marshal_jsonpb.go
// and altered to support Temporal JSONPB implementation. Specifically, the
// JSONPb struct was altered to accept an unmarshaler also and to change the
// marshaler to Temporal's.
```

I would write code way better :-)

internal/temporalgateway/marshal_jsonpb.go
```diff
@@ -0,0 +1,281 @@
...
+//
+// The NewDecoder method returns a DecoderWrapper, so the underlying
+// *json.Decoder methods can be used.
+type JSONPb struct {
```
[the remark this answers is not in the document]
cretz: This is not my code. See comment near top of file.

common/v1/payload_json.go
```diff
@@ -0,0 +1,228 @@
...
+	// We need to try to deserialize into the regular payload first. If it works
+	// and there is metadata _and_ data actually present (or null with a null
+	// metadata encoding), we assume it's a non-shorthand payload. If it fails
+	// (which it will if not an object or there is an unknown field or if
+	// 'metadata' is not string + base64 or if 'data' is not base64), we assume
+	// shorthand. We are ok disallowing unknown fields for payloads here even if
+	// the outer unmarshaler allows them.
+	if gogojsonpb.Unmarshal(bytes.NewReader(b), p) == nil && len(p.Metadata) > 0 {
+		// A raw payload must either have data or a binary/null encoding
+		if len(p.Data) > 0 || string(p.Metadata["encoding"]) == "binary/null" {
```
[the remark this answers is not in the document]
cretz: > if you reference same string more than once, I would like to see them on top of this file in the const block

In this case, these are basically company-wide constants have have no real home. I can move them to _private_ constants in this file, but it makes little sense to me as they are used by two private methods colocated together.

> Similar to DisablePayloadShorthandMetadataKey but these will be really public (exported).

But that is exported for a reason, there is no reason to export these. I only use them in two private methods and never would a user use them (we already have them exported in the Go SDK). I don't think we need to export more consts out of this mostly-generated package.

> Similar [consts](https://github.com/temporalio/sdk-go/blob/master/converter/metadata.go) in Go SDK could just alias them.

Aliasing was a very poor decision. But assuming you mean just have the Go SDK reference a new set of consts here, what benefit is there? If it's just to satisfy the _idea_ that they're shared string literals, I don't think that has enough benefits practically.

> these consts are very well hardcoded (and immutable) not only in this repo but across many repos in entire org.

Exactly. They might as well be part of our proto definitions, but since they aren't we have string literals in the projects that reference them. It is unfortunate that api-go even exists (the SDK and server should be able to own their own proto generation) and luckily other languages aren't burdened with that. If it was reasonable I would not put this logic here, but since this is our "shared Go place for payloads", we need the specialized JSON conversion here I think.

PR #115, opened 2023-08-03
Mitigate injection

jackdawm:
**What changed?**
Replace interpolated github context variables with default variables used more defensively. 

**Why?**
This mitigates an attacker naming a branch a command that could run in our runner's context.

**How did you test it?**
Ran in my own fork and it was fine in my workflow.

**Potential risks**
Somehow not printing the build info in the workflow output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Print build information
-        run: "echo head_ref: ${{ github.head_ref }}, ref: ${{ github.ref }}"
+        run: 'echo head_ref: "$GITHUB_HEAD_REF", ref: "$GITHUB_REF"'
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:


cretz (approved):
Thanks!

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Print build information
-        run: "echo head_ref: ${{ github.head_ref }}, ref: ${{ github.ref }}"
+        run: 'echo head_ref: "$GITHUB_HEAD_REF", ref: "$GITHUB_REF"'
```
cretz, line 13: You may have to set these as environment variables first

dnr: I kind of can't believe that gha has such a security-footgun with no warnings or easy alternatives. I've fixed this in multiple other places. there isn't even a "shell-escape" filter available on substitutions :(

cretz: Ah, sorry! I confused the exec line for the output. Thanks, will merge.

PR #116, opened 2023-08-25
Change update-proto workflow auth

jackdawm:
Instead of using the Commander Data PAT to commit, use the workflow's own secret. Add the permission to commit to this branch in the job definition.

Since this commits but doesn't open a PR, this doesn't need permissions to open or edit PRs. You can unselect that check box and do read-only for the other default options as well.

workflow options in repo

**Why?**
We are trying to move away from PATs and instead have each workflow have its [own minimal permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token).

**How did you test it?**
I ran this in my own fork [here](https://github.com/jackdawm/api-go/actions/runs/5978724793), with the settings I described at the top for default workflow permissions in the repo. It even allowed me to spoof as myself when it was the action, which...might be concerning, actually, since we take this from user inputs, but we've had it that way the whole time.

**Potential risks**
Permissions being wrong and it not updating proto in the desired branch. This can be seen by looking at the output of the workflow run.


cretz (approved):
Makes perfect sense. Will need a server team member to approve as well.

cretz:
> how important it is or is not to allow the person running this workflow to choose a different identity and message for this commit to run as?

Not important. The majority of the time this is triggered from https://github.com/temporalio/api/blob/master/.github/workflows/trigger-api-go-update.yml (so also please make sure that continues to work). Manual dispatch is rare, though I did need it recently.

alexshtin:
I triggered this GHA manually only when I was testing it.

alexshtin (approved):

PR #117, opened 2023-10-03
Replace gogo protobuf with google's protobuf v2 compiler

tdeebswihart:
⚠️ THIS IS A DRAFT PUT UP TO SOLICIT FEEDBACK WHILE THE PROJECT IS IN PROGRESS ⚠️

Our API is currently converted, working, and in-use by our go SDK but I may need to change or add things as I convert our server and other repos. Please provide feedback! There's a lot of work still to do

**What changed?**

gogo/protobuf has been replaced with Google's official go compiler. 

⚠️ I have, for now, deleted our custom JSON coder so I can spread this change across our repos as quickly as possible for testing. I'd like to know ASAP whether we lose measurable performance due to this change. If we do, I will be introducing Vitess' protobuf compiler plugin [vtproto](https://github.com/planetscale/vtprotobuf)

**Why?**

gogo/protobuf has been deprecated for some time and the community is moving on, building new tools (like vtproto) atop google's v2 compiler.

**Breaking changes**

- `*time.Time` in proto structs will now be [timestamppb.Timestamp](https://pkg.go.dev/google.golang.org/protobuf@v1.31.0/types/known/timestamppb#section-documentation)
- `*time.Duration` will now be [durationpb.Duration](https://pkg.go.dev/google.golang.org/protobuf/types/known/durationpb)
- V2-generated structs embed locks, so you cannot dereference them willy-nilly. `go vet` will scream at you about this
- Proto enums will, when formatted to JSON, now be in `SCREAMING_SNAKE_CASE` rather than `PascalCase`. We decided (in discussion with the SDK team) that now was as good a time as any to rip the bandage off

Note that history loading will not be impacted by the JSON changes: I rewrote history loading to dynamically fix incoming history JSON data (like all our other sdks); you can find this code in `internal/temporalhistoryv1/load.go`. That will be used by our Go sdk as well as our CLI when I get there.

**How did you test it?**

All tests for this repo pass as do all SDK tests (except for integration tests, which time out on our official repo on master...)

**Potential risks**

All errors that could arise from this change should be compile-time errors, so your build will be broken if I haven't yet addressed your code. I plan to port _all_ relevant temporal repos as a part of this effort.

In addition the enum-rewriting code (the `fix-enums` makefile target) is rather heinous; if the official go plugin for protobuf ever changes how they name their enums (which they shouldn't) it will break.

**Work to be done**
- [ ] Rewrite the custom JSON coding logic to use protojson


tdeebswihart:
Also closing this to point at a branch. /sigh

PR #119, opened 2023-10-03
Replace gogo protobuf with google's protobuf v2 compiler

tdeebswihart:
**What changed?**

gogo/protobuf has been replaced with Google's official go compiler. 

⚠️ I have, for now, deleted our custom JSON coder so I can spread this change across our repos as quickly as possible for testing. I'd like to know ASAP whether we lose measurable performance due to this change. If we do, I will be introducing Vitess' protobuf compiler plugin [vtproto](https://github.com/planetscale/vtprotobuf)

**Why?**

gogo/protobuf has been deprecated for some time and the community is moving on, building new tools (like vtproto) atop google's v2 compiler.

**Breaking changes**

- `*time.Time` in proto structs will now be [timestamppb.Timestamp](https://pkg.go.dev/google.golang.org/protobuf@v1.31.0/types/known/timestamppb#section-documentation)
- `*time.Duration` will now be [durationpb.Duration](https://pkg.go.dev/google.golang.org/protobuf/types/known/durationpb)
- V2-generated structs embed locks, so you cannot dereference them willy-nilly. `go vet` will scream at you about this
- Proto enums will, when formatted to JSON, now be in `SCREAMING_SNAKE_CASE` rather than `PascalCase`. We decided (in discussion with the SDK team) that now was as good a time as any to rip the bandage off

Note that history loading will not be impacted by the JSON changes: I rewrote history loading to dynamically fix incoming history JSON data (like all our other sdks); you can find this code in `internal/temporalhistoryv1/load.go`. That will be used by our Go sdk as well as our CLI when I get there.

**How did you test it?**

All tests for this repo pass as do all SDK tests (except for integration tests, which time out on our official repo on master...)

**Potential risks**

All errors that could arise from this change should be compile-time errors, so your build will be broken if I haven't yet addressed your code. I plan to port _all_ relevant temporal repos as a part of this effort.

In addition the enum-rewriting code (the `fix-enums` makefile target) is rather heinous; if the official go plugin for protobuf ever changes how they name their enums (which they shouldn't) it will break.

**Work to be done**
- [ ] ~~Rewrite the custom JSON coding logic to use protojson~~ Will be done later and moved to the server repo


batch/v1/message.go-helpers.go
```diff
@@ -0,0 +1,224 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package batch
+
+import "google.golang.org/protobuf/proto"
+
+func (val *BatchOperationInfo) Marshal() ([]byte, error) {
```
cretz, line 27: Can you alter your template to put Godoc on every exported Go function? (even though gogo didn't follow Go norms here)

cmd/proxygenerator/interceptor.go
```diff
@@ -48,7 +48,7 @@ import (
 	"fmt"
 
 	"google.golang.org/grpc"
-	"github.com/gogo/protobuf/proto"
+    "google.golang.org/protobuf/proto"
```
cretz, line 51: `go fmt`

cmd/proxygenerator/service.go
```diff
@@ -79,12 +80,12 @@ func generateService(cfg config) error {
...
 	pkgs, err := packages.Load(conf, "go.temporal.io/api/workflowservice/v1")
 	if err != nil {
-		return err
+		return fmt.Errorf("unable to load workflowservice: %w", err)
 	}
 
 	pkg := pkgs[0]
 	if len(pkg.Errors) > 0 {
-		return fmt.Errorf("unable to load workflowservice: %v", pkg.Errors)
+		return fmt.Errorf("unable to load workflowservice: %w", pkg.Errors)
```
cretz, line 88: Can you use a slice as `%w`?

cmd/proxygenerator/service.go
```diff
@@ -103,19 +104,32 @@ func generateService(cfg config) error {
...
+			fmt.Fprintf(buf, "\nfunc (s *workflowServiceProxyServer) %s(%s) %s {\n", name, strings.Join(paramDecl, ", "), types.TypeString(sig.Results(), qual))
 			fmt.Fprintf(buf, "\treturn s.client.%s(%s)\n", name, strings.Join(params, ", "))
 			fmt.Fprintf(buf, "}\n")
 		}
 	}
 
 	src, err := format.Source(buf.Bytes())
 	if err != nil {
-		return err
+		log.Println(buf.String())
```
cretz, line 131: Leftover debug code? (though I guess harmless to dump code before showing error)

go.mod
```diff
@@ -3,16 +3,14 @@ module go.temporal.io/api
 go 1.16
 
 require (
-	github.com/gogo/googleapis v1.4.1
-	github.com/gogo/protobuf v1.3.2
-	github.com/gogo/status v1.1.1
 	github.com/golang/mock v1.6.0
-	github.com/golang/protobuf v1.5.3
-	github.com/grpc-ecosystem/grpc-gateway v1.16.0
+	github.com/google/go-cmp v0.5.9
```
cretz, line 7: Is this only for tests? I know it may seem a bit much, but this is a really core library to a lot of people's projects, so ideally we can not add any dependencies we don't have to (even if it's a transitive dependency of something we already have, that's better than explicit)

go.mod
```diff
@@ -3,16 +3,14 @@ module go.temporal.io/api
...
+	github.com/grpc-ecosystem/grpc-gateway/v2 v2.18.0
+	github.com/iancoleman/strcase v0.3.0
```
cretz, line 9: Can you inline (poor-man's vendor) this dependency? I would usually never request such a thing, but how this library is used in so many projects coupled with the very limited need for our use is a bit too left-pad-ish.

internal/temporalcommonv1/payload_json.go
cretz, line 1: If you'd like, we can setup some time to hash out some ideas here on how best to support this and gRPC gateway in general (obviously no problem commenting out while you work)

internal/temporalhistoryv1/load.go
```diff
@@ -0,0 +1,162 @@
...
+			historyPathsToFix[fullPath] = strcase.ToScreamingSnake(string(fe.Name()))
+		}
+	}
+
+	if fm := fd.Message(); fm != nil {
+		path = append(path, fd.JSONName())
+		for i := 0; i < fm.Fields().Len(); i++ {
+			cd := fm.Fields().Get(i)
+			// Avoid blowing the stack when a proto may contain itself
+			// I'm looking at you temporal.api.failure.v1.Failure...
```
cretz, line 67: But do you still walk the nested failure cause when fixing? Can you add a test case confirming?

internal/temporalhistoryv1/load.go
```diff
@@ -0,0 +1,162 @@
...
+	switch v := val.(type) {
+	case []interface{}:
+		for i := range v {
+			if vv, ok := v[i].(map[string]interface{}); ok {
+				visitAndFix(vv, path, fixes)
+			}
+		}
+	case map[string]interface{}:
+		for kk, vv := range v {
+			fullPath := fmt.Sprintf("%s.%s", path, kk)
```
cretz, line 87: If you're gonna hardcode `.` here, you might as well do it everywhere and remove the const

internal/temporalhistoryv1/load.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+		}
+	}
+}
+
+// LoadFromJSON deserializes history from a reader of JSON bytes. This does
+// not close the reader if it is closeable. This function is compatible both
+// with the "correct" SCREAMING_SNAKE enums of protojson as well as the old
+// PascalCase enums of Temporal's older history exports.
+func LoadFromJSON(r io.Reader, lastEventID int64) (*History, error) {
```
cretz, line 110: I'd remove the `lastEventID` thing. That's a Go SDK only thing. We can change that SDK to lop events off after unmarshaling over there.

types/duration/helpers.go
cretz, line 1: Do we need these helpers (and therefore a whole new package)? I think if Temporal code needs these tiny things, they can add their own helpers for their own use. I don't think users of this library need these.

cmd/proxygenerator/service.go
tdeebswihart: It's here so the author can debug the error; I had multiple errors that were unsolveable from the message alone, I needed to see the offending buffer in order to fix them.

go.mod
tdeebswihart: It's used for tests because you cannot easily compare protobuf objects; testify's Equal.* method don't work as they use reflection and so will compare the cached serialization state of the protobuf objects (even though its not exported).

I can write a simple reflection-based comparator if we need, but I'm introducing cmp in our downstream repos for the same reason

internal/temporalhistoryv1/load.go
tdeebswihart: That's a good catch: no, it doesn't. I'll add a test then a _separate_ loop to walk the failure tree (much like is done in our .Net api now that I've taken another look at it)

types/duration/helpers.go
tdeebswihart: We can nix them, but it provided a bit smoother of an api than `common.DurationProto`
and the like. We also use these type conversions across our go repos, so it seems a waste to reimplement them separately in every repository that uses our go-api

cretz: > it seems a waste to reimplement them separately in every repository that uses our go-api

I don't think it's that much of a waste. This is the Go way IMO over multiple new packages for few tiny helpers (some of which may not even be needed over the library version).

batch/v1/message.go-helpers.go
tdeebswihart: Done!

internal/temporalhistoryv1/load.go
tdeebswihart: This is done and published in my latest commit. I ended up taking a different approach than .Net that will allow us to handle _any_ type that can contain itself

internal/temporalhistoryv1/load.go
tdeebswihart: Done

internal/temporalhistoryv1/load.go
tdeebswihart: Done!

types/duration/helpers.go
tdeebswihart: Removed.

go.mod
tdeebswihart: Done.

ash2k:
"Kicking the tires" and gogo protobufs are causing a compilation failure in my project. I'm using bazel, so there are a few "gotchas", but... it'd be nice if everything just used the v2 vanilla API. Thank you!

tdeebswihart:
> "Kicking the tires" and gogo protobufs are causing a compilation failure in my project. I'm using bazel, so there are a few "gotchas", but... it'd be nice if everything just used the v2 vanilla API. Thank you!

That's the plan. This effort is to use the vanilla v2 API across all temporal Go repos :)

tdeebswihart:
So it looks like the server encodes and decodes a number of types from JSON, not just history.

I'm going to add a new top-level `jsonpb` package to contain this decoding logic that will also contain the shorthand json encoding logic once I get there (it has to be rewritten anyways)

cretz:
> I'm going to add a new top-level jsonpb package to contain this decoding logic that will also contain the shorthand json encoding logic once I get there (it has to be rewritten anyways)

If you would like, you can just put this on the server. That's the only place I think we need the payload shorthand stuff and probably the only place we need this. I added it here instead of server only because I figured that users might want to encoded/decode shorthand payloads, but happy to change stance on that to keep surface area small. At the least we did this here without adding a new user-facing package to this repo.

tdeebswihart:
> > I'm going to add a new top-level jsonpb package to contain this decoding logic that will also contain the shorthand json encoding logic once I get there (it has to be rewritten anyways)
> 
> If you would like, you can just put this on the server. That's the only place I think we need the payload shorthand stuff and probably the only place we need this. I added it here instead of server only because I figured that users might want to encoded/decode shorthand payloads, but happy to change stance on that to keep surface area small. At the least we did this here without adding a new user-facing package to this repo.

The jsonpb marshaler provided by this repo's proxy module was also used by the SDK; is this unnecessary?

At a bare minimum the decoding logic needs to be used by both the SDK, the server, and perhaps the CLI. I'm open to moving the encoding logic elsewhere

cretz:
> The jsonpb marshaler provided by this repo's proxy module was also used by the SDK; is this unnecessary?

Was it? Or was it `github.com/gogo/protobuf/jsonpb`? I just introduced it for the HTTP API project and I wouldn't have expected anyone to use it but the server's HTTP API impl.

> At a bare minimum the decoding logic needs to be used by both the SDK, the server, and perhaps the CLI

Not for shorthand payloads, that's an HTTP server only thing. I think the SDK and CLI and every other Go library should be able to use normal `protojson` (except in the one case already solved w/ history JSON in the SDK).

tdeebswihart:
> Was it? Or was it github.com/gogo/protobuf/jsonpb? I just introduced it for the HTTP API project and I wouldn't have expected anyone to use it but the server's HTTP API impl.

I checked and it seems I misremembered; the SDK does not look to use our shorthand encoder.

> At a bare minimum the decoding logic needs to be used by both the SDK, the server, and perhaps the CLI

We have two sets of coding logic we care about:

**Shorthand JSON** - I agree that we can place this in the server and remove it from our public API. I'm a fan of this

**All other JSON** - This needs to be supported in multiple repos (if not all our go repos).

cretz:
After off-PR discussion, it was determined that a package can live here for serializing/deserializing old-style proto JSON, though we may prefer a package name like `temporalproto` instead of `jsonpb` (and it can also hold other proto utilities we may have that we've been shoving in the `proxy` package)

workflowservicemock/v1/service_grpc.pb.mock.go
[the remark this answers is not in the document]
ash2k: FWIW I think it makes sense to delete them for the following reasons:
- https://github.com/golang/mock that is used to generate mocks is no longer maintained. https://github.com/uber-go/mock is the recommended replacement (fork).
- Apart from the Uber's fork, there is at least also https://github.com/vektra/mockery, which is also quite popular. So, there is no standard, people use various frameworks.
- It's trivial to generate mocks using one's favorite library.
- Good to have fewer dependencies.

go.mod
tdeebswihart: I can if we really want to. I'll add it to my list

workflowservicemock/v1/service_grpc.pb.mock.go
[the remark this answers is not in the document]
tdeebswihart: Someone should file a ticket for it so we can put it somewhere in the backlog

go.mod
stephanos: FWIW, if we're talking about the risk of the GitHub repo disappearing some day (like left-pad) did; I was told the other day in our Slack that all Go dependencies are cached on Google's proxy and disappearing modules shouldn't be an issue 🤷

workflowservicemock/v1/service_grpc.pb.mock.go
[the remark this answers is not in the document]
ash2k: This project doesn't have an issue tracker. What's the right place? Perhaps https://github.com/temporalio/sdk-go ?

protoc-gen-go-helpers/main.go
cretz, line 1: Can you put this dir under `cmd/`?

temporal-enum-rewriter/main.go
cretz, line 1: Can you put this dir under `cmd/`?

protoc-gen-go-helpers/main.go
tdeebswihart: Yup! I'll move them later today once I have a separate patch out

go.mod
cretz: I did not make the request due to "disappearing modules", I made it to reduce the dependencies on our downstream users. `go.temporal.io/api` is a library is used by hundreds/thousands of users, so there needs to be a good reason to increase its transitive dependency surface. This isn't one IMO.

The main issue with increasing these dependencies is not size or anything, it's versioning. We are now forcing everyone to have v0.3.0 of this library. v0.2.0 and v0.4.0 are not compatible with this version (see https://go.dev/ref/mod#versions) so we have now put unnecessary rules on our users which is bad.

tdeebswihart: Once I have the server's unit tests passing (I'm almost there) I'll just write the two functions (ToSnake, ToCamel) we need myself.

tdeebswihart:
Note: since the HTTP API isn't GA yet I don't plan to include shorthand JSON support in the first pass of all this work. 

I spoke with @yiminc yesterday about this and we can add that in _after_ the initial release

tdeebswihart:
> Note: since the HTTP API isn't GA yet I don't plan to include shorthand JSON support in the first pass of all this work.
> 
> I spoke with @yiminc yesterday about this and we can add that in _after_ the initial release

And we chatted again today. The shorthand JSON support will almost certainly be in the initial release, but I'll reimplement it in the server repo as @cretz mentioned it's the only place we need it

temporalproto/public_methods.go
```diff
@@ -20,27 +20,32 @@
...
-		reflect.Uint64:  reflect.ValueOf(runtime.Uint64),
-		reflect.Uint32:  reflect.ValueOf(runtime.Uint32),
-		reflect.Slice:   reflect.ValueOf(runtime.Bytes),
+var publicMethodRgx = regexp.MustCompile("^[A-Z]")
+
+// PublicMethods calls the provided callback on each method declared as public on the
+// specified object.
+// This prevents the `mustEmbedUnimplementedFooBarBaz` method required by the GRPC v2
+// gateway from polluting our tests.
+func PublicMethods(obj any, cb func(reflect.Method)) {
```
cretz, line 36: How is this used? Must this non-Temporal-non-proto utility be exposed as public API from the Temporal proto library?

workflowservicemock/v1/service_grpc.pb.mock.go
cretz, line 1: Would totally support getting rid of mocks altogether in this repo if you could get buy in from other groups

tdeebswihart: I don't want to do this as part of this work. It's scope is large enough already

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "proto/api"]
 	path = proto/api
-	url = https://github.com/temporalio/api
+	url = https://github.com/tdeebswihart/temporal-api
```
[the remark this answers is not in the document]
tdeebswihart: 💯

cmd/enumrewriter/main.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// All protobuf enums are 32-bit ints thankfully
+type enumVariant struct {
+	Name string
+	// Int32 as string
+	Value string
+}
+
+// return fmt.Sprintf("InvalidTYPENAME(%d)", SELF)
```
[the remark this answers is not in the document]
tdeebswihart: Yeah, whoops. I'll removed it in my next commit

cmd/enumrewriter/main.go
```diff
@@ -0,0 +1,253 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func rewriteFile(fileName string) error {
+	fset := token.NewFileSet()
```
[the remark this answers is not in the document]
tdeebswihart: Heh, yeah. It felt like the most robust way to replace the contents of the String method

temporalproto/decode.go
```diff
@@ -0,0 +1,248 @@
...
+			}
+		}
+	}
+}
+
+// Unmarshal a single proto object from the provided slice of bytes. This function
+// is compatible both with the "correct" SCREAMING_SNAKE enums of protojson as well
+// as the PascalCase enums of earlier releases.
+//
+// This does **not** support slices of proto objects.
```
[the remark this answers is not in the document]
tdeebswihart: I updated the comment in the latest commit, but it means you can't unmarshal []T where T implements proto.Message

temporalproto/decode.go
```diff
@@ -0,0 +1,248 @@
...
+
+	out, err := json.Marshal(msg)
+	if err != nil {
+		return fmt.Errorf("failed to marshal fixed history: %w", err)
+	}
+
+	opts := protojson.UnmarshalOptions{
+		// Ignore unknown fields because if the histroy was generated with a different version of the proto
+		// fields may have been added/removed.
+		DiscardUnknown: true,
```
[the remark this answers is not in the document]
tdeebswihart: I'm updating it to be configurable much like both jsonpb and protojson. I'll plumb it through the SDK and Server and set the DiscardUnknown value wherever we're interacting with history

temporalproto/public_methods.go
tdeebswihart: These should have stayed in the server, whoops.

tdeebswihart: I'd like to keep DeepEqual here at least; it's complex enough that I really don't want us maintaining multiple copies of it (and I'm using it in the CLI's testing)

cretz: Since it is very Temporal proto specific, being in `temporalproto` makes sense for that one.

temporalproto/decode.go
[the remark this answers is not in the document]
cretz: But it's a compiler error to even try to unmarshal to `[]T` anyways

cmd/enumrewriter/main.go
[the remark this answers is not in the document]
cretz: I see we added a README.md just for this one. That's fine, though the other cmds' purposes remain unclear (e.g. `protoc-gen-go-helpers`) but it's probably fine.

temporalproto/decode.go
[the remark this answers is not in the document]
tdeebswihart: I _think_ the API used to take an `any` when I first developed it; you're right that this comment is no longer necessary

cmd/enumrewriter/main.go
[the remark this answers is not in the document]
cretz: :+1: No specific need on any of this, it was just a bit hard to tell their purpose when I started reading their code.

cmd/proxygenerator/interceptor.go
tdeebswihart: What about that ordering is incorrect?

cretz (commented):
Looks good except for newly added dependency on strcase (https://github.com/temporalio/api-go/pull/119/files#r1376163691)

workflowservicemock/v1/service_grpc.pb.mock.go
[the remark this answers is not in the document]
ash2k: Looks like there is already an issue for this https://github.com/temporalio/sdk-go/issues/61.

PR #122, opened 2023-10-11
Bump Go to 1.20

stephanos:
**What changed?**

Bumped Go to 1.20.

**Why?**

`grpc` dependency requires 1.19+ now and without bumping Go up, changes to this repo are blocked:

```
# google.golang.org/grpc
Error: /home/runner/go/pkg/mod/google.golang.org/grpc@v1.58.3/server.go:2096:14: undefined: atomic.Int64
note: module requires Go 1.19
```


Makefile
```diff
@@ -7,7 +7,7 @@ install: grpc-install mockgen-install goimports-install update-proto
 proto: grpc goimports proxy grpc-mock copyright
 
 # Update submodule and compile proto files.
-update-proto: update-proto-submodule proto update-dependencies gomodtidy
+update-proto: update-proto-submodule gomodtidy proto update-dependencies
```
stephanos, line 10: `gomodtidy` should come before `proto`!

tdeebswihart: No it should come after. This will cause errors as we're generating packages we require to run tidy

tdeebswihart (changes requested):
You can't put gomodtidy before proto generation; I found that out the hard way on my branch

stephanos:
@tdeebswihart okay, reverted. Out of curiosity, why not?

tdeebswihart:
> @tdeebswihart okay, reverted. Out of curiosity, why not?

It blew up on me last time I tried, but perhaps it was something else. I'm not having trouble swapping them now /shrug

stephanos:
@tdeebswihart anything blocking a merge now?

go.mod
```diff
@@ -10,9 +10,18 @@ require (
 	github.com/golang/protobuf v1.5.3
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/stretchr/testify v1.8.4
+	google.golang.org/genproto/googleapis/api v0.0.0-20231009173412-8bfb1ae86b6c
```
cretz, line 13: I assume this is just a side effect of changing version?

go.mod
```diff
@@ -10,9 +10,18 @@ require (
...
+	google.golang.org/grpc v1.58.3
```
cretz, line 14: Was confused at first by title of PR, but this is really "Bump Go to 1.20 and upgrade gRPC library"

Can you give more details on the need to upgrade gRPC? (not that I disagree, just want to understand)

enums/v1/workflow.pb.go
cretz, line 1: Can we not update generated code in this PR? If the CI tool that auto-gens code on `api` repo merge updates these files, so be it.

go.mod
stephanos: Yes, the `go mod tidy` did all of these. Which I see happen on most commits as a "side effect".

go.mod
stephanos: I can revert the library updates if you want and only bump Go, but the thing is since `go mod tidy` is part of the Makefile target to generate the files, the next person will inevitably update this. That's actually what led me here to having to bump the Go version.

enums/v1/workflow.pb.go
stephanos: Oh, I see. I can revert those 👍

go.mod
stephanos: I wonder if we want to decouple "generate files" from "bump versions"? It's a little odd that they are so tightly coupled.

This is the part of the Makefile that I'm referring to:
```
# Compile proto files.
proto: grpc goimports proxy grpc-mock copyright

# Update submodule and compile proto files.
update-proto: update-proto-submodule proto update-dependencies gomodtidy
```
Notice how `update-dependencies` is part of `update-proto`.

enums/v1/workflow.pb.go
stephanos: @cretz update proposal to the README.me for the next person: https://github.com/temporalio/api-go/pull/125

go.mod
stephanos: Nevermind, I understand how the workflow of this and https://github.com/temporalio/api work now; I've reverted any library updates and made this PR focus on the Go version bump!

cretz (approved):

PR #128, opened 2023-11-02
Handle wrapped ServiceErrors in Convert

dnr:
**What changed?**
Make `Convert` handle wrapped `ServiceError`s

**Why?**
We shouldn't lose any more specific error info (i.e. grpc code) if it's present at all in the error.

**How did you test it?**
unit test

**Potential risks**
This code path is a little more expensive, if we end up using it a lot of the time that's a slight waste.

### DIFF
--- a/serviceerror/convert.go
+++ b/serviceerror/convert.go
@@ -43,7 +43,7 @@ func ToStatus(err error) *status.Status {
 		return svcerr.Status()
 	}
 	// err does not implement ServiceError directly, but check if it wraps it.
-	// This path does more copying so prefer to return a ServiceError directly if possible.
+	// This path does more allocation so prefer to return a ServiceError directly if possible.
 	var svcerr ServiceError
 	if errors.As(err, &svcerr) {
 		s := svcerr.Status().Proto()
--- a/serviceerror/convert.go
+++ b/serviceerror/convert.go
@@ -42,6 +42,14 @@ func ToStatus(err error) *status.Status {
 	if svcerr, ok := err.(ServiceError); ok {
 		return svcerr.Status()
 	}
+	// err does not implement ServiceError directly, but check if it wraps it.
+	// This path does more copying so prefer to return a ServiceError directly if possible.
+	var svcerr ServiceError
+	if errors.As(err, &svcerr) {
+		s := svcerr.Status().Proto()
+		s.Message = err.Error() // don't lose the wrapped message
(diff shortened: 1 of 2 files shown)


serviceerror/convert.go
```diff
@@ -42,6 +42,14 @@ func ToStatus(err error) *status.Status {
 	if svcerr, ok := err.(ServiceError); ok {
 		return svcerr.Status()
 	}
+	// err does not implement ServiceError directly, but check if it wraps it.
+	// This path does more copying so prefer to return a ServiceError directly if possible.
+	var svcerr ServiceError
+	if errors.As(err, &svcerr) {
+		s := svcerr.Status().Proto()
```
MichaelSnowden, line 49: Just double-checking, this does a copy right? In any case, can we verify in the test that the original message is unmodified?

dnr: I'm not sure I would say it does a "copy". It asks the ServiceError to serialize itself into a google.golang.org/genproto/googleapis/rpc/status.Status (proto message). The it replaces one field in the proto message and constructs a google.golang.org/grpc/status.Status from that. Those both involve copying some fields, but we're moving between different types. Specifically, there's only ever one ServiceError.

yycptt: I think the underlying implementation of Proto() does a Clone() so original msg should be unmodified.

yycptt (approved):

serviceerror/convert.go
```diff
@@ -42,6 +42,14 @@ func ToStatus(err error) *status.Status {
 	if svcerr, ok := err.(ServiceError); ok {
 		return svcerr.Status()
 	}
+	// err does not implement ServiceError directly, but check if it wraps it.
+	// This path does more allocation so prefer to return a ServiceError directly if possible.
+	var svcerr ServiceError
+	if errors.As(err, &svcerr) {
```
cretz, line 48: I support this change because this is what https://pkg.go.dev/google.golang.org/grpc/status does. But we should have never had a `ServiceError` IMO because `GRPCStatus()` should have been the method and existing gRPC status package should have been good enough. But I guess that ship sailed long ago.

Technically it's a backwards incompatible change, but from an SDK POV there are no obvious user-facing times where this will accidentally return a status where it wouldn't have otherwise because user-facing deserialized error hierarchy only doesn't contain this type at any depth (it's not a failure type).

PR #134, opened 2023-11-28
Port protojson coding stack to support shorthand HTTP payloads

tdeebswihart:
## 🚨 REVIEWER NOTES
- you can ignore the marshal and unmarshal tests, those are ported
- this won't compile as the necessary `internal/protojson` code is ported in https://github.com/temporalio/api-go/pull/135 (stacked pr) to minimize the review load
- I've added comments at the relevant bits to review in the protojson code I've adapted. You're welcome to review the rest of those files but I didn't write them

## Description
**What changed?**
I replaced the janky JSON fixup code I wrote previously with an altered version of the official protojson coding stack.

This provides both shorthand JSON payloads for our HTTP API as well as efficient support for both new and old-style JSON enums.

**Why?**
I replaced the janky JSON fixup code I wrote previously with an altered version of the official protojson coding stack. Our old shorthand payload support was built upon gogo/protobuf's json coding stack, which won't work with google/protobuf.

**How did you test it?**
See all the tests I added

**Potential risks**
At worst this will break our JSON encoding, but I'd be surprised if it did.


temporalproto/json_marshal.go
```diff
@@ -0,0 +1,386 @@
...
+		return marshal(e, m)
+	}
+	if jsu, ok := m.Interface().(JSONPBMaybeMarshaler); ok {
+		if handled, err := jsu.MaybeMarshalJSONPB(e.opts.Metadata, e.Encoder); handled {
+			if err != nil {
+				return fmt.Errorf("MaybeMarshalJSONPB impl for %T failed: %w", jsu, err)
+			}
+			return nil
+		}
+	}
```
tdeebswihart, line 241: This is the only real alteration to the JSON marshalling code

temporalproto/json_marshal.go
```diff
@@ -0,0 +1,386 @@
...
+	var err error
+	if typeURL != "" {
+		// We manually marshal the synthetic @type field here as we cannot create a synthetic FieldDescriptor.
+		// We cannot implement the ProtoInternal method as it takes an internal-only type as a parameter...
+		if err := e.WriteName("@type"); err != nil {
+			return err
+		}
+		if e.WriteString(typeURL) != nil {
+			return errors.InvalidUTF8("@type")
+		}
```
tdeebswihart, line 263: This is also a difference from the official code, but I ported all the non-legacy tests and we pass so 🤷

temporalproto/json_marshal.go
```diff
@@ -0,0 +1,386 @@
...
+	case protoreflect.EnumKind:
+		if fd.Enum().FullName() == genid.NullValue_enum_fullname {
+			e.WriteNull()
+		} else {
+			desc := fd.Enum().Values().ByNumber(val.Enum())
+			if e.opts.UseEnumNumbers || desc == nil {
+				e.WriteInt(int64(val.Enum()))
+			} else {
+				e.WriteString(string(desc.Name()))
+			}
```
tdeebswihart, line 342: If we decide, for some reason, _not_ to change our JSON enums to SCREAMING_SNAKE then this is where we'd emit the pascalCase form

temporalproto/json_unmarshal.go
```diff
@@ -0,0 +1,701 @@
...
+	return errors.New(head+f, x...)
+}
+
+// unmarshalMessage unmarshals a message into the given protoreflect.Message.
+func (d decoder) unmarshalMessage(m protoreflect.Message, skipTypeURL bool) error {
+	if jsu, ok := m.Interface().(JSONPBMaybeUnmarshaler); ok {
+		if handled, err := jsu.MaybeUnmarshalJSONPB(d.opts.Metadata, d.Decoder); handled || err != nil {
+			return err
+		}
+	}
```
tdeebswihart, line 151: This is the main difference from the official protojson coder

Makefile
```diff
@@ -63,9 +63,11 @@ go-grpc: clean .go-helpers-installed $(PROTO_OUT)
 fix-path: go-grpc
 	mv -f $(PROTO_OUT)/temporal/api/* $(PROTO_OUT) && rm -rf $(PROTO_OUT)/temporal
 
+# Copy the payload helpers
 copy-helpers:
-	# Copy the payload helpers
+	chmod +w $(PROTO_OUT)/common/v1/payload_json.go 2>/dev/null || true
 	cp $(PROTO_OUT)/internal/temporalcommonv1/payload_json.go $(PROTO_OUT)/common/v1/
+	chmod -w $(PROTO_OUT)/common/v1/payload_json.go
```
tdeebswihart, line 70: I remove write permissions after copying so I stop accidentally editing the wrong file

temporalproto/compatibility_test.go
```diff
@@ -136,22 +135,22 @@ func TestUnmarshalJSON(t *testing.T) {
...
 func TestUnmarshalJSON_Compatible(t *testing.T) {
 	t.Parallel()
-	u := temporalproto.JSONUnmarshaller{
+	u := temporalproto.JSONUnmarshalOptions{
 		DiscardUnknown: true,
 	}
 	// Ensure both new and old enums deserialize the same way
 	var oldHist, newHist historypb.History
-	require.NoError(t, u.Unmarshal([]byte(newEnums), &oldHist))
+	require.NoError(t, u.Unmarshal([]byte(oldEnums), &oldHist))
```
tdeebswihart, line 143: This is a test bug I fixed

temporalproto/json_unmarshal.go
```diff
@@ -0,0 +1,715 @@
...
+		s := tok.ParsedString()
+		if enumVal := fd.Enum().Values().ByName(protoreflect.Name(s)); enumVal != nil {
+			return protoreflect.ValueOfEnum(enumVal.Number()), true
+		}
+
+		// If not found we'll assume it's old-style and prepend TYPE_PREFIX_
+		s = strcase.ToScreamingSnake(fmt.Sprintf("%s_%s", string(fd.Enum().Name()), s))
+		if enumVal := fd.Enum().Values().ByName(protoreflect.Name(s)); enumVal != nil {
+			return protoreflect.ValueOfEnum(enumVal.Number()), true
+		}
```
tdeebswihart, line 527: This allows us to support either enum variant

go.mod
```diff
@@ -4,10 +4,11 @@ go 1.20
 
 require (
 	github.com/golang/mock v1.6.0
+	github.com/google/go-cmp v0.6.0
```
cretz, line 7: Is this new dependency really required for everyone using this library? Nobody downstream will be able to update to 0.7.0 when it comes out before we do it because we've forced this on them (Go mod rules say minor of 0-ver not compatible w/ next minor of 0-ver).

internal/temporalcommonv1/payload_json.go
```diff
@@ -22,205 +22,587 @@
...
+}
+
+var _ temporalproto.JSONPBMaybeMarshaler = (*Payload)(nil)
+var _ temporalproto.JSONPBMaybeMarshaler = (*Payloads)(nil)
+var _ temporalproto.JSONPBMaybeUnmarshaler = (*Payload)(nil)
+var _ temporalproto.JSONPBMaybeUnmarshaler = (*Payloads)(nil)
 
 // !!! This file is copied from internal/temporalcommonv1 to common/v1.
 // !!! DO NOT EDIT at common/v1/payload_json.go.
+func marshalSingular(enc *json.Encoder, value interface{}) error {
```
cretz, line 61: Did you hand-write this marshaling or is it from another library? Why not put this stuff the vendored area somewhere internal with the rest of the vendored code?

temporalproto/json_marshal.go
cretz, line 1: Can we just move most of this vendored code to somewhere internal? Why expose all of this to users and clutter up our package we're using for other things with all of these files that serve a single support-special-JSON-interface purpose? Can we discuss exactly what API we want exposed to users?

internal/temporalcommonv1/payload_json.go
```diff
@@ -22,205 +22,587 @@
...
+
+const (
+	payloadMetadataKey      = "metadata"
+	payloadDataKey          = "data"
+	shorthandMessageTypeKey = "_protoMessageType"
+)
+
+var logfunc func(msg string, args ...any) (int, error) = fmt.Printf
+
+func SetLogfunc(f func(msg string, args ...any) (int, error)) {
```
cretz, line 50: Should not add package-level function to our public API for this purpose, and definitely should default it to printf, and even if you had to have it, user-facing API needs Godoc

temporalproto/maybe_marshaler.go
```diff
@@ -0,0 +1,44 @@
...
+
+// JSONPBMaybeMarshaler is implemented by any proto struct that wants to
+// customize optional Temporal-specific JSON conversion.
+type JSONPBMaybeMarshaler interface {
+	// MaybeMarshalJSONPB is for formatting the proto message as JSON. If the
+	// "handled" result value is false, "b" and "err" are ignored and the default
+	// proto JSON behavior occurs. currIndent is the current prefix depth but
+	// should not be applied if m.Indent is empty.
+	MaybeMarshalJSONPB(meta map[string]interface{}, enc *json.Encoder) (handled bool, err error)
+}
```
cretz, line 35:
```suggestion
type ProtoJSONMaybeMarshaler interface {
	// MaybeMarshalProtoJSON is for formatting the proto message as JSON. If the
	// "handled" result value is false, "b" and "err" are ignored and the default
	// proto JSON behavior occurs. currIndent is the current prefix depth but
	// should not be applied if m.Indent is empty.
	MaybeMarshalProtoJSON(meta map[string]interface{}, enc *json.Encoder) (handled bool, err error)
}
```

Probably should change this since "jsonpb" is no longer a term used by modern Go proto JSON libraries

cretz (commented):
Did a first high-level round of review. Basically I think we should not clutter other stuff up with this JSON proto problem if we can and we should keep user-facing API as limited as is reasonable (so the interfaces and the ability to marshal/unmarshal fine, but all of the vendored code I think deserves to not live next to Temporal code)

go.mod
tdeebswihart: This is used by the official protojson code for the most part, so our downstreams are already stuck with it. I simply added it here so the test was better:
```
$ rg -t go -l go-cmp
internal/protojson/json/decode_test.go
internal/protojson/json/encode_test.go
internal/protojson/order/order_test.go
internal/temporalcommonv1/payload_json_test.go
temporalproto/json_marshal_test.go
temporalproto/compatibility_test.go
```

internal/temporalcommonv1/payload_json.go
tdeebswihart: This is hand-written so there's nothing to vendor. I mean, I can if we want but it's not used elsewhere.

temporalproto/json_marshal.go
tdeebswihart: We could, but the public API will need to be the same and it's already pretty small. We need MarshalOptions exposed so the server and SDK can configure it, and we need UnmarshalOptions as we configure it when deserializing history  from JSON.

We could expose functions that wrap these (and keep them internal), setting them to specific options

internal/temporalcommonv1/payload_json.go
tdeebswihart: Whoops, I meant to delete this. This was for testing specifically

temporalproto/maybe_marshaler.go
tdeebswihart: I'm happy with that. I'll go update it

go.mod
cretz: Our downstreams can update Go proto independently if/when Go proto updates and also updates this library. But us now taking an explicit dependency on it means they can't, we have to update this dependency. I'd really like to keep this a transitive dependency instead of an explicit one. Do we have to use it?

internal/temporalcommonv1/payload_json.go
cretz: Ah, ok, I see now and understand why you wrote this. I do think this is probably too general-purpose to be in a generated file/package, but if you don't want yet another package, ok. I will now review this as newly written code.

temporalproto/json_marshal.go
cretz: IMO it's a bit rough to have several files written by Temporal and several files written by not-Temporal in the same package. I do like the idea of non-Temporal code to be out of this package (and we expose our forms which can be simple translations), but my opinion isn't that strong.

temporalproto/maybe_marshaler.go
cretz: Per off-PR discussion, I think we settled on good naming/exposure

go.mod
tdeebswihart: If we rewrite all the protojson tests that use it then no. I _could_ just delete those tests as well, but I'd like to be sure this continues working.

I'll see how much pain rewriting those checks would be

temporalproto/json_marshal.go
tdeebswihart: Yeah, I agree with you there. What I think I'll do is move the protojson code into internal/protojson, then expose a slim options struct that will configure the internal-only options struct when used.

That's _probably_ the best we can do

go.mod
cretz: Thanks! I know it seems a bit overly cautious, but definitely feels safer to not infect our downstream users with our test-only dependency

tdeebswihart: I rewrote them. You'll see the effect in the other PR

cretz (approved):
A gorgeously tiny user-facing footprint, thanks!

PR #137, opened 2023-11-30
Retract version 1.26.1

tdeebswihart:
**What changed?**
I've marked version 1.26.1 as [retracted](https://go.dev/ref/mod#go-mod-file-retract) so that `go get -u` will not longer upgrade to it. It will still be available if you manually specify it, meaning our downstream repos require no changes.

**Why?**
This was released before we were entirely ready and it caused some user development workflows to break

**How did you test it?**
N/A

**Potential risks**


tdeebswihart:
Once this is merged (as we discussed in slack) I'll delete the github release.

Quinn-With-Two-Ns (approved):

cretz (approved):

PR #139, opened 2023-12-07
Add protogen build tool

tdeebswihart:
**What changed?**
I've added the `protogen` build tool to manage compiling and post-processing protobuf-generated code.

**Why?**

We do a lot of work to post-process protobuf-generated code for our purposes and it's not sustainably to rewrite and maintain it separately for each repo when we're always doing the same thing.

This tool combines all of the following into a single build tool:

- proto compilation
- enum post-processing (renaming enum consts and rewriting String methods)
- stripping plugin versions from the generated files
  - Alpine's version of proto is a full major version behind brew's, and some of our repos will fail CI if you differ from the alpine-generated code

**How did you test it?**
This tool is used in the PR stacked on this one that rebuilds the code as well as TODO in the server repo. I expect to add it to the CDS and saas-temporal repos as well once it's merged.

**Potential risks**

None. If I've broken something in another repo I'll fix it once I add it to that repo


cmd/protogen/main.go
```diff
@@ -0,0 +1,345 @@
...
+	"go/token"
+	"io/fs"
+	"os"
+	"os/exec"
+	"os/signal"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"go.temporal.io/api/internal/errgroup"
```
cretz, line 43: Meh, probably not worth having/vendoring this as there is probably very little benefit to concurrently doing any of this. Just keep it simple. Surely not worth a whole new package of vendored code in this repo for such likely-unnecessary use. If you must have this, I would support making this cmd be its own `go.mod` with `golang.org/x/sync/errgroup` imported.

cmd/protogen/main.go
```diff
@@ -0,0 +1,345 @@
...
+
+	"go.temporal.io/api/internal/protogen/enum"
+	"go.temporal.io/api/internal/protogen/version"
```
cretz, line 46: I'd say just inline these in here, the only place they'll ever be used. This is a bit too much unnecessary package/code sprawl for what should just be a simple utility.

cmd/protogen/main.go
tdeebswihart: I started by doing that but was unable to `go install` it from other repos, unfortunately. Perhaps I misconfigured something; you can see that in one of the earlier commits

cretz: Hrmm, how did you install something from a fork before it was even merged? We know this works, because it's how people install our [`workflowcheck` tool](https://github.com/temporalio/sdk-go/tree/master/contrib/tools/workflowcheck) which is a child module of the primary repo.

Regardless, probably don't need a separate module or errgroup

tdeebswihart: As for concurrency: I'm a fan of fast builds. I'll gather some simple timing metrics of the concurrent implementation and compare that against a serial one so we can decide whether it's worth the effort.

tdeebswihart: The current concurrent implementation run against the server repo is around a full second faster per run.

Concurrent:
```
$ hyperfine --warmup 3 'make protoc'
Benchmark 1: make protoc
  Time (mean ± σ):      2.021 s ±  0.140 s    [User: 1.969 s, System: 1.768 s]
  Range (min … max):    1.816 s …  2.249 s    10 runs
```

Serial implementation:
```
$ hyperfine --warmup 3 'make protoc'
Benchmark 1: make protoc
  Time (mean ± σ):      2.955 s ±  0.100 s    [User: 1.966 s, System: 1.555 s]
  Range (min … max):    2.769 s …  3.108 s    10 runs
```

cretz: I don't think you'd find anyone that isn't a fan of fast builds. I'm sure we could pour over the code and eek out some more wins if we were micro-optimizing, but I question the overall value, _especially_ since this is to replace the existing makefile that was probably serial.

tdeebswihart: Sure. I'd prefer to optimize it further, but I'd rather have something maintainable in first. I've removed all concurrency with the latest commit

cretz (approved):
:+1: Looks like a good, simple approach. (I think if string manip instead of AST could simplify more, but this LGTM)

PR #143, opened 2024-01-03
Port DOS protections to our protojson fork

tdeebswihart:
**What changed?**

I ported over the [DOS fixes](https://github.com/protocolbuffers/protobuf-go/commit/bfcd6476a38e41247d6bb43dc8f00b23ec9fffc2) from the official protobuf repository

**Why?**

Two DOS bugs were fixed by golang/protobuf recently:

1. https://github.com/golang/protobuf/issues/1583
2. https://github.com/golang/protobuf/issues/1584

**How did you test it?**

I ported over their tests for the issue

**Potential risks**

None,


cretz (approved):
In a perfect world we'd maintain just a set of patches on top of existing lib, but that is annoying.

internal/protojson/well_known_types.go
```diff
@@ -309,48 +309,25 @@ Loop:
...
-				if err := d.skipJSONValue(); err != nil {
-					return err
-				}
+		switch tok.Kind() {
+		case json.ObjectClose, json.ArrayClose:
+			open--
+		case json.ObjectOpen, json.ArrayOpen:
+			open++
+			if open > d.opts.RecursionLimit {
+				return errors.New("exceeded max recursion depth")
```
rodrigozhou, line 324: You're not doing recursion anymore, so does it matter to track the depth?

tdeebswihart: This is copied verbatim from the official codebase

tdeebswihart:
> In a perfect world we'd maintain just a set of patches on top of existing lib, but that is annoying.

We'd need to fork the whole protobuf stack to do so, which would suck. We could do that and periodically rebase our changes on the upstream, but that's a lot of work

PR #145, opened 2024-01-23
Fix marshalling of shorthand maps

tdeebswihart:
**What changed?**

1. Map values may not be singular items but full values
2. Empty maps should be marshalled as `{}`, not `null`

This fixes temporalio/api#349

**Why?**

Empty maps should always have been marshalled as empty objects, this was a bug in our code. While here I realized a deeper bug, where maps would only support simple values (strings, numbers, etc) but not embedded maps, lists, etc.

**How did you test it?**

New unit test

**Potential risks**

None


cretz (approved):
Thanks!

PR #146, opened 2024-02-05
Generate OpenAPI v2 and v3 specs

tdeebswihart:
**What changed?**

We now generate openapi [v2 (swagger)](https://swagger.io/specification/v2/) and [v3](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md) specs as a part of proto compilation.

**Why?**

So that we can serve these on our HTTP API

**How did you test it?**

Build and run https://github.com/temporalio/temporal/pull/5393 (make start-dependencies + make start)
```
$ curl localhost:7243/api/v1/swagger.json
$ curl localhost:7243/api/v1/openapi.yaml
```

**Potential risks**


Quinn-With-Two-Ns:
Should this be in the next server RC?

tdeebswihart:
> Should this be in the next server RC?

I don't believe so, we're aiming at a point release for server 1.23.

tdeebswihart:
> Should this be in the next server RC?

No. We plan to release this in a point release

openapi/spec.go
```diff
@@ -0,0 +1,36 @@
...
+
+// The OpenAPI package embeds compressed copies of Temporal's OpenAPI v2 (swagger) and v3
+// specifications.
+//
+// Do not import this package unless you need to: it will bloat your binary's size by
+// around 83KB
+package openapi
+
+// OpenAPI v2 "swagger" specification of the Temporal HTTP API
+// go:embed openapiv2.swagger.json
```
cretz, line 31: I thought that go:embed didn't work for libs hence the generated files

cmd/encode-openapi-spec/main.go
```diff
@@ -0,0 +1,109 @@
...
+type templateInput struct {
+	Version int
+	Format  string
+	Spec    string
+}
+
+const tmpl = `package openapi
+
+// OpenAPIv{{.Version}}Spec contains a gzip-compressed {{.Format}} file specifying the Temporal HTTP API
+var OpenAPIv{{.Version}}Spec = {{.Spec}}`
```
cretz, line 47:
```suggestion
// OpenAPIV{{.Version}}Spec contains a gzip-compressed {{.Format}} file specifying the Temporal HTTP API
var OpenAPIV{{.Version}}Spec = {{.Spec}}`
```

openapi/openapiv3.go
cretz, line 1: I wonder if this package should be a child of `temporalproto` instead of a top-level package? Not that it matters much. (to clarify, if you do want to do this, should still be its own package)

openapi/spec.go
tdeebswihart: This file was from an earlier revision of this work, let me remove it

openapi/openapiv3.go
tdeebswihart: I'd be fine hiding it under `temporalproto/openapi` if we want to reduce the number of top-level modules. It'll only be used by the HTTP api, so there's no reason to pollute the module root

cmd/encode-openapi-spec/main.go
```diff
@@ -0,0 +1,109 @@
...
+type templateInput struct {
+	Version int
+	Format  string
+	Spec    string
+}
+
+const tmpl = `package openapi
+
+// OpenAPIV{{.Version}}Spec contains a gzip-compressed {{.Format}} file specifying the Temporal HTTP API
+var OpenAPIV{{.Version}}Spec = {{.Spec}}`
```
cretz, line 47:
```suggestion
var OpenAPIV{{.Version}}{{.Lang}}Spec = {{.Spec}}`
```

Sorry for not making this observation earlier, but I think we need to put `YAML` or `JSON` in the name. I was under the impression at least one of those OpenAPI versions supports both.

openapi/swagger.go
cretz, line 1: Can this file be removed?

cmd/encode-openapi-spec/main.go
tdeebswihart: Ah good catch, I'll add that

cretz (approved):

PR #148, opened 2024-03-01
Fix disabling utf-8 validation for nested types

dnr:
**What changed?**
Fix #147 for nested types

**Why?**
See #147 

**How did you test it?**
Also added unit test (just for rewrite code itself, not for effects)


tdeebswihart (approved):

cmd/protogen/disable_utf8.go
```diff
@@ -92,8 +104,10 @@ func (v *disableUtf8Validation) visit(n ast.Node) bool {
 			newElts := make([]ast.Expr, len(newArr))
 			for i, v := range newArr {
 				newElts[i] = &ast.BasicLit{
-					Kind:  token.INT,
-					Value: fmt.Sprintf("%#02x", v),
+					// steal positions from original list to preserve newlines
```
dnr, line 107: I read through a bunch of the formatter/printer.. it looks like adding newlines in exactly the right places (and specifically adding more than there were before) is quite a lot more work. this is a pretty easy way to just get _some_ newlines in there.

cretz (approved):
LGTM, but would personally like to see one test here and one in Go SDK:

* A test somewhere outside of the generator (i.e. in `temporalproto` package or something) that confirms reusing one or more of our proto types that string fields, map string keys, repeated strings, and a string in a nested message all marshal and unmarshal non-utf-8 as expected
* A test in Go SDK that confirms end-to-end that you can set something like a header key and/or a client identity to a non-utf8 string and that it works all the way through server and back on history.

(but neither are blocking)

tdeebswihart:
> LGTM, but would personally like to see one test here and one in Go SDK:
> 
>     * A test somewhere here but outside of the generator (i.e. in `temporalproto` package or something) that confirms, reusing one or more of our proto types, that string fields, map string keys, repeated strings, and a string in a nested message all marshal and unmarshal non-utf-8 as expected
> 
>     * A test in Go SDK that confirms end-to-end that you can set something like a header key and/or a client identity to a non-utf8 string and that it works all the way through server and back on history.
> 
> 
> (but neither are blocking)

I plan to add to add tests all the way through our stack (sdk, server, etc) to make sure that everywhere we regenerate/use protos that we support this behavior

test/invalidutf8_test.go
cretz, line 1: Can this be moved under `temporalproto` or something and not a new top-level folder?

PR #149, opened 2024-03-05
NamespaceInfo.Data should be a map to bytes

tdeebswihart:
**What changed?**
The namespace data map now has `bytes` values rather than `string`.

**Why?**

We embed binary data (cast to strings) here in a few places, and that can cause google/protobuf to yell at us for having string data that's invalid UTF-8. Protobuf strings and bytes are wire-compatible

**How did you test it?**

**Potential risks**
This will change all code generated by protoc plugins to use the language's type for raw bytes. I have paired Go API and server PRs that will go along with this change


tdeebswihart:
We've decided to axe this and migrate the relevant fields internally

PR #150, opened 2024-03-06
Update golang.org/x/tools dependency in proxygenerator

rodrigozhou:
**What changed?**
Update golang.org/x/tools dependency in proxygenerator to latest version.

**Why?**
Make it work with Go 1.21 and Go 1.22

**How did you test it?**
`make proto` works with Go 1.22

**Potential risks**


tdeebswihart (approved):

cretz (approved):
Thanks! Would also support a matrix of Go versions in the CI at some point (though surprised all Go software isn't always backwards compatible, but I guess technically there are internal changes on what `/x/tools` leverages).

PR #154, opened 2024-03-22
Fix shorthand encoding of Payloads

dnr:
**What changed?**
Marshal/unmarshal empty `Payloads` as `[]` in shorthand encoding (also accept `null` on unmarshal).
Handle the case where one or more `Payload` within `Payloads` can't be encoded as shorthand.

**Why?**
These cases were producing invalid json.

**How did you test it?**
unit tests

### DIFF
--- a/internal/temporalcommonv1/payload_json_test.go
+++ b/internal/temporalcommonv1/payload_json_test.go
@@ -102,6 +102,30 @@ var tests = []struct {
 		},
 		Data: []byte(`{"greeting":{"name":{}}}`),
 	},
+}, {
+	name:          "empty payloads",
+	longformJSON:  `{}`,
+	shorthandJSON: `null`,
+	pb:            &common.Payloads{},
+}, {
+	name:          "empty payloads with non-nil slice",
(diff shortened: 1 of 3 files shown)


internal/temporalcommonv1/payload_json_test.go
```diff
@@ -102,6 +102,30 @@ var tests = []struct {
 		},
 		Data: []byte(`{"greeting":{"name":{}}}`),
 	},
+}, {
+	name:          "empty payloads",
+	longformJSON:  `{}`,
+	shorthandJSON: `null`,
```
cretz, line 108: Hrmm, arguably this could be `[]`

cretz (approved):
Looks great! Let's let @tdeebswihart review too before merging.

internal/temporalcommonv1/payload_json_test.go
tdeebswihart: An empty array seems "correct" as it is an empty array, but we should support parsing either

PR #155, opened 2024-03-22
Clean cloud folder in Makefile

cretz:
**What changed?**

In preparation for https://github.com/temporalio/api/pull/378, the `clean` target needs to work with a subdir like `cloud`, so since we have no non-generated code in `cloud`, we just delete the entire directory on `clean`


Quinn-With-Two-Ns:
hmm  shouldn't we require approval to merge?

cretz:
Closing as the approach of putting cloud protos with the rest of protos has been rejected at this time

PR #156, opened 2024-03-26
Adding ResourceScope to ResourceExhaustedFailure

prathyushpv:
**What changed?**
Adding a new field ResourceScope in ResourceExhaustedFailure.

This PR depends on https://github.com/temporalio/api/pull/382. 

**Why?**
We need to differentiate whether a request has been rate limited because of namespace limit or system limit.
This will help us better understand why requests are rate limited and which limits to adjust.

**How did you test it?**

**Potential risks**
None


serviceerror/resourceExhausted.go
```diff
@@ -34,15 +34,17 @@ type (
...
 		Cause   enumspb.ResourceExhaustedCause
+		Scope   enumspb.ResourceScope
 		Message string
 		st      *status.Status
 	}
 )
 
 // NewResourceExhausted returns new ResourceExhausted error.
-func NewResourceExhausted(cause enumspb.ResourceExhaustedCause, message string) error {
+func NewResourceExhausted(cause enumspb.ResourceExhaustedCause, scope enumspb.ResourceScope, message string) error {
```
cretz, line 44: While we may think this is only used by us, this is public API at https://pkg.go.dev/go.temporal.io/api/serviceerror and we shouldn't break compatibility IMO. Can we either have a `NewResourceExhaustedWithOptions` or request that those building this with a scope create the struct directly?

prathyushpv: Okay! I thought only server and go-sdk use this function. Let me make that change.

prathyushpv: I have removed this argument from this function. It is better to create this struct directly.

cretz: Thanks! Will approve when api repo PR merged and regen occurs and is merged into this branch

prathyushpv: Thanks @cretz! I have merge api changes..

cretz (approved):

PR #158, opened 2024-04-17
Add MultiOperation errors

stephanos:
> [!IMPORTANT]  
> This is based on the yet un-merged API in change https://github.com/temporalio/api/pull/398

**What changed?**

Added two new error types, both for MultiOperationExecution.

The errors behavior was documented in https://github.com/temporalio/api/pull/367.

**Why?**

To return these errors from the Server ([see PR](https://github.com/temporalio/temporal/pull/5745)).

**How did you test it?**

Add new tests and [Server implementation PR](https://github.com/temporalio/temporal/pull/5745).

**Potential risks**

None


serviceerror/convert_test.go
```diff
@@ -98,6 +98,37 @@ func TestToStatus_NotServiceError(t *testing.T) {
 	require.Len(t, st1.Details(), 0)
 }
 
+func TestToStatus_MultiOp(t *testing.T) {
```
cretz, line 101: What about `FromStatus` so Go SDK users can see these errors?

serviceerror/multiOpAborted.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	failurepb "go.temporal.io/api/failure/v1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var errMultiOperationAborted = MultiOperationAborted{}
+
+type MultiOperationAborted struct{}
```
cretz, line 11: IMO should add Godoc to all exported items (unsure if we do elsewhere though)

cretz, line 11:
```suggestion
type MultiOperationAborted struct {
	Message string
	st      *status.Status
}
```

And allow message to be passed in constructor and make the pointer to this implement error instead of this whole. Benefits:

* This is used by SDKs too and should not swallow the
* Pointer receiver can carry more future information without copying regularly
* Status as a field means every invocation of `Status()` doesn't have to make a new one (it's usually meant to be a cheap getter IME)

If you want this to be a fixed package-level var on server side with a fixed string message and a fixed status, no prob. But don't bake it into the type used in SDK and server.

All of this applies to the other error too. Design for SDK and server use instead of just server use.

serviceerror/convert_test.go
stephanos: 👍 I didn't realize that the Go SDK is calling `FromStatus`.

serviceerror/multiOpAborted.go
stephanos: I made all these choices intentionally because this particular error is different from the others in that it's meant to be a singleton/sentinel error, essentially. I don't see how any of these benefits will materialize for this particular error type in practice, but fair enough, I can align it with all the other types for consistency.

serviceerror/convert_test.go
cretz: :+1: This library is (transitively) our most used library by users :grin:

serviceerror/multiOpAborted.go
cretz: I think when you align this object for client-side use, it'll make more sense to have these fields

serviceerror/convert.go
```diff
@@ -71,7 +72,21 @@ func FromStatus(st *status.Status) error {
 		return nil
 	}
 
-	// Simple case. `st.Code()` to `serviceerror` is one to one mapping and there are no error details.
+	errDetails := extractErrorDetails(st)
```
stephanos, line 75: The error parsing had to be changed so that it always parses the error details first. This adds an extra step for status codes such as Unavailable and Canceled, but is most likely negligible.

serviceerror/multiOpAborted.go
stephanos: Okay, done!

serviceerror/multiOpAborted.go
```diff
@@ -0,0 +1,64 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package serviceerror
+
+import (
+	failurepb "go.temporal.io/api/failure/v1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
cretz, line 28: Mentioned in API repo PR, but this type of gRPC dependency isn't going to be available in some other SDK languages and we can't really force it on users, so we have to have our own generated proto of the status proto which can clash when they _do_ include the gRPC dependency for their other needs.

stephanos: https://github.com/temporalio/api/pull/398 has been updated to use its own copy of `google.rpc.Status`.

serviceerror/convert.go
```diff
@@ -87,20 +107,18 @@ func FromStatus(st *status.Status) error {
...
-	// then errDetails will be of type `error` with corresponding error inside.
-	// This error is ignored and `serviceerror` is built using `st.Code()` only.
-
-	switch st.Code() {
+	case codes.Aborted:
+		switch errDetails.(type) {
+		case *failure.MultiOperationExecutionAborted:
+			return newMultiOperationAborted(st)
+		default:
+			return st.Err()
```
alexshtin, line 120: I would do same as on line 190, or add generic `AbortedError` with empty details.

serviceerror/multiOpAborted.go
alexshtin, line 1: `multi_op_aborted.go` please. And someone need to rename all other files in this package.

serviceerror/multiOpAborted.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// MultiOperationAborted represents an aborted operation from a MultiOperationExecution.
+type MultiOperationAborted struct {
+	Message string
+	st      *status.Status
+}
+
+// NewMultiOperationAbortedError returns MultiOperationAborted.
+func NewMultiOperationAbortedError(message string) error {
```
alexshtin, line 38: You need to be consistent in using `Error` suffix. I agree on not using `Failure` suffix for proto message and I would prefer to use `Error` suffix for error because it is essentially a error in go world. Or remove `Error()` method.

Makefile
```diff
@@ -41,7 +41,7 @@ $(PROTO_OUT):
 ##### git submodule for proto files #####
 update-proto-submodule:
 	printf $(COLOR) "Update proto-submodule..."
-	git -c protocol.file.allow=always submodule update --init --force --remote $(PROTO_ROOT)
+	#git -c protocol.file.allow=always submodule update --init --force --remote $(PROTO_ROOT)
```
alexshtin, line 44: Don't forget to uncomment this.

alexshtin (approved):

cretz (commented):
I was intentionally waiting on API repo merge before giving approval here, but it looks like it was merged w/ no SDK approval because our branch protections and codeowners don't require it. I wonder if we should change this repo to be like the `api` repo w/ the restrictions there.

stephanos:
@cretz
Sorry, I should have pinged you (or someone else from SDK) again. If there are any concerns with this PR in particular, please let me know.

I agree, it's probably a good idea to require SDK approval here. I looked over the recent PRs and most - but not all - had SDK approval (but most didn't have a Server approval).

PR #164, opened 2024-05-29
Update GHA update-proto.yml

rodrigozhou:
**What changed?**
- Update actions versions.
- Add optional input to update protos of a specific branch
- Use Github token instead of the secrets `COMMANDER_DATA_TOKEN`
- Fix bug when the commit message contains double-quotes.

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/.github/workflows/update-proto.yml
+++ b/.github/workflows/update-proto.yml
@@ -3,6 +3,10 @@ name: "Update Proto"
 on:
   workflow_dispatch:
     inputs:
+      branch:
+        description: "Branch in api-go repo to update protos (default: master)"
+        required: false
+        default: master
       commit_author:
         description: "Commit author username"
         required: true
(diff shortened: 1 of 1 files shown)


.github/workflows/update-proto.yml
```diff
@@ -3,6 +3,10 @@ name: "Update Proto"
 on:
   workflow_dispatch:
     inputs:
+      branch:
+        description: "Branch to update protos (default: master)"
```
cretz, line 7: This confused me at first, I thought it may be the branch of the API repo we were updating but it's actually the branch of this repo we're pushing to. Unsure if clarification needed here.

cretz (approved):
I think the last attempt at this failed and we had to revert, hopefully this one works

.github/workflows/update-proto.yml
rodrigozhou: Updated description to be crystal clear. :)

PR #167, opened 2024-06-04
Forward metadata in proxy

cretz:
**What changed?**

Today headers set by client do not flow through the proxy. This will now forward them by default.

### DIFF
--- a/proxy/service_util.go
+++ b/proxy/service_util.go
@@ -1,0 +1,38 @@
+package proxy
+
+import (
+	"context"
+
+	"google.golang.org/grpc/metadata"
+)
+
+func (s *workflowServiceProxyServer) reqCtx(ctx context.Context) context.Context {
+	if s.disableHeaderForwarding {
+		return ctx
+	}
+
+	// Copy incoming header to outgoing if not already present in outgoing. We
+	// have confirmed in gRPC that incoming is a copy so we can mutate it.
+	incoming, _ := metadata.FromIncomingContext(ctx)
+
+	// Remove common headers and if there's nothing left, return early
+	incoming.Delete("user-agent")
+	incoming.Delete(":authority")
+	incoming.Delete("content-type")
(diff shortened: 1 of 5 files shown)


proxy/service_util.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func (s *workflowServiceProxyServer) reqCtx(ctx context.Context) context.Context {
+	if s.disableHeaderForwarding {
+		return ctx
+	}
+
+	// Copy incoming header to outgoing if not already present in outgoing. We
+	// have confirmed in gRPC that incoming is a copy so we can mutate it.
+	incoming, _ := metadata.FromIncomingContext(ctx)
```
tdeebswihart, line 16: Error check?

proxy/service_util.go
```diff
@@ -0,0 +1,38 @@
...
+
+	// Remove common headers and if there's nothing left, return early
+	incoming.Delete("user-agent")
+	incoming.Delete(":authority")
```
tdeebswihart, line 20: Is this supposed to have a colon?

proxy/service_util.go
cretz: This isn't an error, this is whether it was present, and we don't care really in our case

tdeebswihart (approved):

cretz:
This should not be merged until _after_ a Go SDK release is made with https://github.com/temporalio/sdk-go/pull/1502. We did a type conversion in the Go SDK on options set here so we cannot release an API library version with this updated or SDK users will get compile errors.

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
@cretz https://github.com/temporalio/sdk-go/pull/1502 is included in `v1.27.0`

cretz:
Thanks! I will go ahead and merge this after CI passes.

PR #169, opened 2024-07-05
Fix marshaling of null field values

dandavison:
Fixes https://github.com/temporalio/cli/issues/607

Fix handling of nil value in custom JSON Marshaler.

### Evidence this is correct
PR contains a new test case that fails without the fix

**Potential risks**
Could break anything that uses this JSON Marshaler, e.g. printing payloads by CLI.

### DIFF
--- a/internal/temporalcommonv1/payload_json_test.go
+++ b/internal/temporalcommonv1/payload_json_test.go
@@ -84,6 +84,16 @@ var tests = []struct {
 		},
 	},
 }, {
+	name:          "json/plain with object with a null value",
+	longformJSON:  `{"metadata":{"encoding":"anNvbi9wbGFpbg=="},"data":"eyJteWtleSI6bnVsbH0="}`,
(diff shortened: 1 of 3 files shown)


internal/temporalcommonv1/payload_json_test.go
```diff
@@ -83,6 +83,16 @@ var tests = []struct {
 			},
 		},
 	},
+}, {
+	name:          "json/plain with object with a null value",
+	longformJSON:  `{"metadata":{"encoding":"anNvbi9wbGFpbg=="},"data":"eyJteWtleSI6bnVsbH0="}`,
```
dandavison, line 88:
```
$ echo -n '{"mykey":null}' | base64
eyJteWtleSI6bnVsbH0=
```

tdeebswihart (approved):

dandavison:
This had already been fixed at https://github.com/temporalio/api-go/pull/159.

PR #170, opened 2024-07-19
GHA to create release

rodrigozhou:
**What changed?**
GHA to create, publish and delete releases.
The workflow can only be triggered by another workflow (from `api` repo).

**Why?**
Automation.

**How did you test it?**

**Potential risks**

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,4 +1,4 @@
-name: "Create a release"
+name: "Create release"
 
 on:
   workflow_call:
@@ -7,9 +7,12 @@ on:
         description: "Ref to be tagged"
         required: true
         type: string
-        default: master
       tag:
         description: "Tag for new version (v1.23.4)"
+        required: true
+        type: string
+      api_commit_sha:
+        description: "api repo commit sha"
         required: true
         type: string
       base_tag:
@@ -24,7 +27,7 @@ on:
 
 jobs:
   create-release:
-    name: "Create a release"
+    name: "Create release"
     runs-on: ubuntu-latest
 
     steps:
@@ -40,7 +43,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: temporalio/api-go
-          ref: ${{ github.event.inputs.ref }}
+          ref: ${{ inputs.ref }}
           token: ${{ steps.generate_token.outputs.token }}
           persist-credentials: true
           fetch-depth: 0
@@ -49,22 +52,29 @@ jobs:
 
       - name: Validate input
         env:
-          REF: ${{ github.event.inputs.ref }}
-          TAG: ${{ github.event.inputs.tag }}
-          BASE_TAG: ${{ github.event.inputs.base_tag }}
+          REF: ${{ inputs.ref }}
+          TAG: ${{ inputs.tag }}
+          API_COMMIT_SHA: ${{ inputs.api_commit_sha }}
+          BASE_TAG: ${{ inputs.base_tag }}
         run: |
           if ! [[ "${TAG}" =~ ^v.* ]]; then
             echo "::error::Tag is not prefixed with 'v'"
             exit 1
           fi
 
-          if [[ -n "$(git tag -l "$TAG")" && "$(git rev-parse "$TAG")" != "$(git rev-parse HEAD)" ]]; then
-            echo "::error::Tag already exists and it doesn't reference $REF"
+          if [[ -n "$(git tag -l "$TAG")" ]]; then
+            echo "::error::Tag already exists"
             exit 1
           fi
 
           if [[ -z "$BASE_TAG" || -z "$(git tag -l "$BASE_TAG")" ]]; then
             echo "::error::Base tag not specified or does not exist"
+            exit 1
+          fi
+
+          API_GO_API_COMMIT_SHA=$(git rev-parse HEAD:proto/api)
+          if [[ "${API_GO_API_COMMIT_SHA}" != "${API_COMMIT_SHA}" ]]; then
+            echo "::error::api-go ref ${API_GO_COMMIT_SHA} does not reference api ref ${API_COMMIT_SHA}."
             exit 1
           fi
 
@@ -76,9 +86,10 @@ jobs:
       - name: Create release
         env:
           GH_TOKEN: ${{ steps.generate_token.outputs.token }}
-          REF: ${{ github.event.inputs.ref }}
-          TAG: ${{ github.event.inputs.tag }}
-          BASE_TAG: ${{ github.event.inputs.base_tag }}
+          REF: ${{ inputs.ref }}
+          TAG: ${{ inputs.tag }}
+          API_COMMIT_SHA: ${{ inputs.api_commit_sha }}
+          BASE_TAG: ${{ inputs.base_tag }}
         run: |
-          gh repo set-default ${{ github.repository }}
-          gh release create "$TAG" --target "$REF" --latest --generate-notes --notes-start-tag "$BASE_TAG"
+          gh repo set-default temporalio/api-go
+          gh release create "$TAG" --title "${TAG}-${API_COMMIT_SHA}" --target "$REF" --latest --generate-notes --notes-start-tag "$BASE_TAG" --draft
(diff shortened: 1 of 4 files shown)


.github/workflows/update-repo-deps.yml
cretz, line 1: Blindly bumping api-go is not recommended. Firstly, rarely do users use this without SDK, and the SDK needs to be vetted on the API bump. Secondly, it makes assumptions about where people's go.mod's live which is not necessarily accurate either. If a certain repo wants some automation to bump a certain dependency it has, that automation should be in that repo IMO.

cretz:
Can you give some background on the savings here on starting a GH workflow vs publishing release in GH?

.github/workflows/update-repo-deps.yml
rodrigozhou: Removed

rodrigozhou:
@cretz Remove potentially some mistakes when releasing a new tag like forgetting to set as latest. Also, with this workflow, we can trigger it whenever the `api` repo has a new release so they can be in sync.

cretz:
> Also, with this workflow, we can trigger it whenever the api repo has a new release so they can be in sync.

To confirm, today's approach is:

* Click release in GH for api
* Click release in GH for api-go

But this new approach is:

* Click run workflow in GH for api

So you save a single click, but now you have two more GH workflows to maintain and the user can still mistakenly click release in api/api-go? If we're concerned about users reading release process documentation and may not properly make one of the two releases, how can we know that they will read the documentation to know not to use the GH release feature but instead manual workflow dispatch instead? Are we saving much or just adding GH workflows to maintain?

rodrigozhou:
It's also to eliminate the human error of forgetting to release `api-go` after `api` is released.

Also, we can further remove the ability of manually triggering the release in `api-go`, and only be possible from the `api` repo (basically change `workflow_dispatch` to `workflow_call`). Although I'd do in separate PR to make sure it's working as it is first.

cretz (approved):
I am a bit concerned about the maintenance of these two workflows for such an incredibly simple thing, but if y'all are resolving to maintain them and are now in full control of the release process of this and the API repo, ok.

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,104 @@
...
+        description: "Create release and set as latest"
+        type: boolean
+        default: true
+      base_tag:
+        description: "Base tag to generate commit list for release notes"
+        required: false
+        type: string
+
+jobs:
+  create-tag:
```
josh-berry, line 25: Please confirm sdk-go builds correctly with the commit to be tagged before tagging the release. This is IMO an essential part of this automation, because without it, we are just making it easier to push out something broken to users.

josh-berry (dismissed):

.github/workflows/create-tag.yml
cretz: To clarify, not `sdk-go@master` but `sdk-go@latest`. We need to make sure our latest API tag works with our latest Go SDK tag. Granted on the Go SDK side, we should update the `api-go` before each release so the inverse is tested as well. I will circulate that requirement separately.

rodrigozhou: The check is being done in the `api` workflow, see https://github.com/temporalio/api/pull/437
I changed this workflow to `workflow_call` event so it can't be triggered manually.

cretz (approved):

.github/workflows/create-tag.yml
josh-berry: 👍 Dropping my objection to merge then.

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Prepare new version string
+        id: new_version
+        env:
+          TAG: ${{ github.event.inputs.tag }}
+        run: |
+          if [[ "${TAG}" =~ ^v.* ]]; then
+              echo "tag=${TAG}" >> "$GITHUB_OUTPUT"
+          else
+              echo "tag=v${TAG}" >> "$GITHUB_OUTPUT"
```
bergundy, line 67: I would just fail and expect a proper tag name to be provided.

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,106 @@
...
+      ref:
+        description: "Ref to be tagged"
+        required: true
+        type: string
+        default: master
+      tag:
+        description: "Tag for new version (v1.23.4)"
+        required: true
+        type: string
+      create_release:
```
bergundy, line 15: Part of the problem is that we tag without creating a release, this shouldn't be optional.

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,106 @@
...
+          fi
+
+      - name: Validate input
+        env:
+          REF: ${{ github.event.inputs.ref }}
+          TAG: ${{ steps.new_version.outputs.tag }}
+          CREATE_RELEASE: ${{ github.event.inputs.create_release }}
+          BASE_TAG: ${{ github.event.inputs.base_tag }}
+        run: |
+          if [[ -n "$(git tag -l "$TAG")" && "$(git rev-parse "$TAG")" != "$(git rev-parse HEAD)" ]]; then
```
bergundy, line 77: I think we want to be even stricter here and make sure we're not skipping versions. At that point, I'd write a Go script to do the validation though.

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Create release
+        if: ${{ github.event.inputs.create_release == true || github.event.inputs.create_release == 'true' }}
+        env:
+          GH_TOKEN: ${{ steps.generate_token.outputs.token }}
+          TAG: ${{ steps.new_version.outputs.tag }}
+          BASE_TAG: ${{ github.event.inputs.base_tag }}
+        run: |
+          gh repo set-default ${{ github.repository }}
+          gh release create "$TAG" --verify-tag --latest --generate-notes --notes-start-tag "$BASE_TAG"
```
bergundy, line 106: Let's create the tag here and publish a draft. I would rather have a human in the loop to verify and edit release notes.

.github/workflows/create-tag.yml
rodrigozhou: But I wonder if there will be a case in which you want to release a patch for a previous version.

.github/workflows/create-tag.yml
rodrigozhou: I wonder if making it a draft will re-introduce the human error of forgetting to publish the release.

.github/workflows/create-tag.yml
bergundy: I think patches are acceptable but I've never seen us issue them.

.github/workflows/create-tag.yml
bergundy: No, because there won't be a tag until the release is published AFAICT.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,84 @@
...
+        with:
+          repository: temporalio/api-go
+          ref: ${{ github.event.inputs.ref }}
+          token: ${{ steps.generate_token.outputs.token }}
+          persist-credentials: true
+          fetch-depth: 0
+          fetch-tags: true
+          submodules: true
+
+      - name: Validate input
```
bergundy, line 50: We should also validate that the version given here matches the `api` submodule version. I think we should verify that there is a release or maybe a draft referencing the `api` submodule commit with the same version given as input to this workflow.

rodrigozhou: Done, I also added the same validation in the `api` workflow before the release drafts are created.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,95 @@
+name: "Create release"
+
+on:
+  workflow_call:
```
bergundy, line 4: We should discourage users from triggering this directly, could you add a note that will show up in the GH UI to encourage users to trigger the `api` release instead?

bergundy (approved):

.github/workflows/create-release.yml
rodrigozhou: This is different from `workflow_dispatch`. This is to allow calling from another workflow.

PR #171, opened 2024-08-09
fix makefile

ychebotarev:
**What changed?**
Add "$(GOBIN)" when calling goimport and mockgen.
Add updating proto api-cloud submodule

**Why?**
Otherwise "make" doesn't work.

**How did you test it?**
by running "make" untill it succeed


cmd/proxygenerator/go.mod
cretz, line 1: I don't think it's a big deal, but can these go.mod/sum files be left unchanged in this PR? We can upgrade/update some of this, but seems unrelated to this PR.

Makefile
```diff
@@ -38,8 +38,10 @@ $(PROTO_OUT):
 
 ##### git submodule for proto files #####
 update-proto-submodule:
-	printf $(COLOR) "Update proto-submodule..."
+	printf $(COLOR) "Update api proto-submodule...\n"
 	git -c protocol.file.allow=always submodule update --init --force --remote $(PROTO_ROOT)
+	printf $(COLOR) "Update cloud proto-submodule...\n"
+	git -c protocol.file.allow=always submodule update --init --force --remote $(PROTO_CLOUD_ROOT)
```
cretz, line 44: Assuming this actually pulls down latest submodule changes, I don't think we want this for the cloud submodule. I think we want that submodule to be updated manually as it is wanted (ideally the repo goes away soon anyways).

Makefile
```diff
@@ -80,9 +82,9 @@ copy-helpers:
 
 grpc-mock:
 	printf $(COLOR) "Generate gRPC mocks..."
-	mockgen -package operatorservicemock -source operatorservice/v1/service_grpc.pb.go -destination operatorservicemock/v1/service_grpc.pb.mock.go
-	mockgen -package workflowservicemock -source workflowservice/v1/service_grpc.pb.go -destination workflowservicemock/v1/service_grpc.pb.mock.go
-	mockgen -package cloudservicemock -source cloud/cloudservice/v1/service_grpc.pb.go -destination cloud/cloudservicemock/v1/service_grpc.pb.mock.go
+	$(GOBIN)/mockgen -package operatorservicemock -source operatorservice/v1/service_grpc.pb.go -destination operatorservicemock/v1/service_grpc.pb.mock.go
```
alexshtin, line 85: I don't like this. We don't have in other `Makefile`s and I would like them to be more or less consistent across repos. [Makefile](https://github.com/temporalio/temporal/blob/main/Makefile) in server repo uses slightly different approach. I am ok, if you copy it from there.

cmd/proxygenerator/go.mod
```diff
@@ -1,6 +1,8 @@
 module go.temporal.io/api/cmd/proxygenerator
 
-go 1.20
+go 1.21
+
+toolchain go1.22.3
```
alexshtin, line 5: While I agree with @cretz, you can replace this with just:
```suggestion
go 1.22.3
```

cretz: No, that would be even worse. That would tell people they need at least 1.22.3 to use this. This libraries is the most depended upon user library in Temporal and we should at least support all non-EOL Go versions (but ideally we don't change any of this unless there is a good reason, and even if there is a good reason, not sure it should be in a makefile-related PR).

ychebotarev:
ok, I will close this one.

PR #172, opened 2024-08-16
Make sure the export package is visitable

Sushisource:
Go version auto-updated itself, seemingly because of dependencies.

**What changed?**
Add the top level type of the export package to the visit walking mechanics

**Why?**
So users can visit payloads on these types

**How did you test it?**
Upgraded test

**Potential risks**
Nada

Closes https://github.com/temporalio/sdk-go/issues/1566


proxy/interceptor_test.go
```diff
@@ -27,6 +27,8 @@ package proxy
 import (
 	"context"
 	"fmt"
+	"go.temporal.io/api/export/v1"
+	"go.temporal.io/api/history/v1"
```
cretz, line 31: Go fmt I would expect would fix this to move non-stdlib imports to later, but no biggie

cretz (approved):

PR #174, opened 2024-08-28
Bump Go version for CI builds to 1.22

lina-temporal:
**What changed?**
- Bumped Go version used for CI builds to 1.21

**Why?**
- Our automated merge builds [began failing](https://github.com/temporalio/api-go/actions/runs/10604768987/job/29392179469)
- [gRPC release 1.66](https://github.com/grpc/grpc-go/tree/v1.66.0?tab=readme-ov-file), which the auto commit updated the package to, supports 1.22 and 1.23.  

**How did you test it?**
- Pending

**Potential risks**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
-          go-version: "1.20"
+          go-version: "1.21"
       - name: Check
         run: make check test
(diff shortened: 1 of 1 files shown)


Shivs11 (approved):
lgtm!

cretz (approved):
Ug, I think this points out a problem we have with our automated merge script. For whatever reason we are updating _every_ dependency blindly when we update protos in that GH workflow. This forces every user we have to upgrade their versions even if we/they don't need to.

Approving this, but am going to bring this up to team. I think forced automatic dependency upgrades are a bad thing for what is effectively Temporal's most used library by users.

PR #176, opened 2024-09-04
Remove automatic dependency upgrade on proto update

cretz:
**What changed?**

Previously we automatically updated all dependencies on every proto submodule update GitHub workflow (triggered by merge in `api` repo). This means that we are forcing users to upgrade things like gRPC unnecessarily. This removes that step from the proto submodule update.


tdeebswihart (approved):
Thank you!

PR #177, opened 2024-09-05
Visit "Any" types in payload visitor

cretz:
**What changed?**

* Added `WellKnownAnyVisitor` to `VisitPayloadsOptions` with default for visiting any
* Updated generator to make sure "Any" types are visited

Closes https://github.com/temporalio/sdk-go/issues/1622

### DIFF
--- a/cmd/proxygenerator/interceptor.go
+++ b/cmd/proxygenerator/interceptor.go
@@ -67,6 +67,9 @@ type VisitPayloadsOptions struct {
 	Visitor func(*VisitPayloadsContext, []*common.Payload) ([]*common.Payload, error)
 	// Don't visit search attribute payloads.
 	SkipSearchAttributes bool
+	// Will be called for each Any encountered. If not set, the default is to recurse into the Any
+	// object, unmarshal it, visit, and re-marshal it always (even if there are no changes).
+	WellKnownAnyVisitor func(*VisitPayloadsContext, *anypb.Any) error
 }
 
 // VisitPayloads calls the options.Visitor function for every Payload proto within msg.
@@ -159,6 +162,25 @@ func NewFailureVisitorInterceptor(options FailureVisitorInterceptorOptions) (grp
 	}, nil
 }
 
+func (o *VisitPayloadsOptions) defaultWellKnownAnyVisitor(ctx *VisitPayloadsContext, p *anypb.Any) error {
+	child, err := p.UnmarshalNew()
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal any: %w", err)
+	}
+	// We choose to visit and re-marshal always instead of cloning, visiting,
+	// and checking if anything changed before re-marshaling. It is assumed the
+	// clone + equality check is not much cheaper than re-marshal.
+	if err := visitPayloads(ctx, o, p, child); err != nil {
+		return err
+	}
+	// Confirmed this replaces both Any fields on non-error, there is nothing
+	// left over
+	if err := p.MarshalFrom(child); err != nil {
+		return fmt.Errorf("failed to marshal any: %w", err)
+	}
+	return nil
+}
+
 func visitPayload(
 	ctx *VisitPayloadsContext,
 	options *VisitPayloadsOptions,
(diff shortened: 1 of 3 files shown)


cmd/proxygenerator/interceptor.go
```diff
@@ -193,15 +215,15 @@ func visitPayloads(
 				if o == nil { continue }
 				no, err := visitPayload(ctx, options, parent, o)
 				if err != nil { return err }
-                objs[i] = no
```
cretz, line 196: This file has some whitespace changes (this was using spaces which was looking bad even though the final generated source was always go fmt'd anyways)

cmd/proxygenerator/interceptor.go
```diff
@@ -159,6 +162,25 @@ func NewFailureVisitorInterceptor(options FailureVisitorInterceptorOptions) (grp
...
 }
 
+func (o *VisitPayloadsOptions) defaultWellKnownAnyVisitor(ctx *VisitPayloadsContext, p *anypb.Any) error {
+	child, err := p.UnmarshalNew()
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal any: %w", err)
+	}
+	// We choose to visit and re-marshal always instead of cloning, visiting,
+	// and checking if anything changed before re-marshaling. It is assumed the
+	// clone + equality check is not much cheaper than re-marshal.
```
cretz, line 172: Does everyone agree with this assumption?

cmd/proxygenerator/interceptor.go
```diff
@@ -159,6 +162,25 @@ func NewFailureVisitorInterceptor(options FailureVisitorInterceptorOptions) (grp
 	}, nil
 }
 
+func (o *VisitPayloadsOptions) defaultWellKnownAnyVisitor(ctx *VisitPayloadsContext, p *anypb.Any) error {
```
cretz, line 165: There could be a need in the future to expose this for users to delegate to, but we will wait until that need to expose it

Quinn-With-Two-Ns (approved):

cmd/proxygenerator/interceptor.go
tdeebswihart: Can we benchmark that assumption? I'd like to see numbers personally

cretz: We can get numbers. Can you define the parameters of the benchmarks for your numbers? Do you want the only way we use "any" today (update requests that have barely anything in them) or benchmarking this with more possibilities in "any"? And do you have recommend sizes of payloads? Meaning do you want to benchmark an update with 4K byte inputs or just 5 bytes (or a bunch of other values)?

Also, it should be noted that this utility is mostly used for payload manipulation which almost always mutates the payloads no matter what making this check-never-mutated piece moot.

tdeebswihart: If that's the context around this assumption, then we should encode that in the code so it's obvious next time looks at an undocumented assumption :)

cretz: It was not meant to be the context around the assumption. I definitely made the assumption regardless of that context. I can provide benchmark numbers, I just need some context around which numbers/scenarios you want to see.

cretz: Interesting results here. So here are some basic benchmarks:

```go
func Benchmark_AnyCloneEqualCheck(b *testing.B) {
	anyMsg, _ := anypb.New(&update.Request{Input: &update.Input{Args: &common.Payloads{
		Payloads: []*common.Payload{{Data: []byte("orig-val")}},
	}}})
	for i := 0; i < b.N; i++ {
		child, _ := anyMsg.UnmarshalNew()
		childOrig := proto.Clone(child)
		// Do visitor stuff...
		proto.Equal(childOrig, child)
	}
}

func Benchmark_AnyMarshalReMarshal(b *testing.B) {
	anyMsg, _ := anypb.New(&update.Request{Input: &update.Input{Args: &common.Payloads{
		Payloads: []*common.Payload{{Data: []byte("orig-val")}},
	}}})
	for i := 0; i < b.N; i++ {
		child, err := anyMsg.UnmarshalNew()
		require.NoError(b, err)
		// Do visitor stuff...
		_ = anyMsg.MarshalFrom(child)
	}
}
```

Running this (on Windows at least) shows that the current unmarshal-remarshal way is at least 3x faster every time I try:

```
Benchmark_AnyCloneEqualCheck
Benchmark_AnyCloneEqualCheck-20           464596              2538 ns/op
Benchmark_AnyMarshalReMarshal
Benchmark_AnyMarshalReMarshal-20         1717684               810.9 ns/op
```

PR #178, opened 2024-09-06
Switch to uber gomock

yycptt:
**What changed?**
- Switch to uber gomock
- Will do the same change in server repo as well.

**Why?**
- google gomock is no longer maintained and doesn't support generics
- Generated mocks are used in server repo as well. and we are switching server repo to uber gomock.

**How did you test it?**

**Potential risks**


stephanos (approved):

yycptt:
Will generate mocks inside server repo.

PR #182, opened 2024-10-14
Visit OperatorService methods in interceptor

bergundy:
**What changed?**

- Visit OperatorService methods in interceptor
- Reenable `generatorcheck` Makefile target

**Why?**

There are payloads in the Nexus EndpointSpec description that need to be visited for completeness.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ test: copy-helpers
 
 generatorcheck:
 	printf $(COLOR) "Check generated code is not stale..."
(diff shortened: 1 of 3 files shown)


cretz (approved):
It is a bit rough that I'm allowed to encrypt endpoint descriptions if I self host but not in cloud. Arguably we should not have made self-hosted descriptions payloads.

bergundy:
We want to allow encryption in Cloud too. There's a planned breaking change to switch the description field to a payload once we figure out how cloud / and OSS protos are going to be combined.

PR #188, opened 2024-11-12
Try rehydrating NamespaceNotActive from Unavailable errors

bergundy:
We want to change the status code to be Unavailable, this is the first step in making this change.


yiminc (approved):

bergundy:
Decided to create a new error instead of changing the semantics of the current one.

PR #190, opened 2024-11-13
Sync to latest API Cloud to include Export

alice-yin:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**


bergundy:
Looks like a lot of unrelated files were changed in this PR. I would guess due to using a different version of the protoc plugin. Can you confirm that this is intentional?

cretz:
We really need a step in CI that checks generated code didn't change

alice-yin:
@cretz @bergundy 

I only added the cloud related files. It has a lot of files as we haven't synced changes in this repo for a while. But if I'm running `make proto` in the whole repo, a lot of other files got pulled in. Example pr: https://github.com/temporalio/api-go/pull/189/files 

Do we want to do that as well? Is it from previous pr?

cretz:
> But if I'm running make proto in the whole repo, a lot of other files got pulled in

Looking at the diff on those, it looks like the copyright headers are removed and your version of protoc may be different than what was here before. I'd have to set aside time to investigate, but in theory if your running of the `make` command ends up with anything different than what's already in `master` for non-cloud things, something may be different with your environment or how it is run. `master` is automatically kept up to date with the API repo.

alice-yin:
yeah. you're right. Running `make` helps me update a lot of dependencies. Previously I'm running `make proto` @bergundy @cretz this is up for review now.

cretz (approved):
Looks good to me. Will you also be doing this for the cloud git submodule inside https://github.com/temporalio/sdk-java/tree/master/temporal-serviceclient/src/main and the cloud git subtree inside https://github.com/temporalio/sdk-core/tree/master/sdk-core-protos/protos (see https://github.com/temporalio/sdk-core?tab=readme-ov-file#proto-files)?

We definitely don't want to favor Go as if it is more important than other languages.

bergundy (approved):

PR #194, opened 2024-11-21
Add NamespaceUnavailable serviceerror

bergundy:
Continue the work I started in https://github.com/temporalio/api/pull/481.

**Why?**

To differentiate between a namespace that's temporary unavailable and not active.
This new failure will be returned with a gRPC unavailable error and should be retried by clients automatically.


serviceerror/namespace_unavailable.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+type (
+	// NamespaceUnavailable is returned by the service when a request addresses a namespace that is unavailable. For
+	// example, when a namespace is in the process of failing over between clusters. This is a transient error that
+	// should be automatically retried by clients.
+	NamespaceUnavailable struct {
+		Namespace string
+		// Note that other service errors have a private status.Status field, there's no compelling reason to do
+		// copy that pattern here. The status can and should be computed on the fly in case the
```
cretz, line 41: Is there a compelling reason to _not_ copy the pattern? _Technically_ one reason that copying the pattern is that the status object could contain additional details. But that's not likely to happen. Still, I see no good reason not to keep the status object as we always have.

bergundy: I'm actually thinking of taking it to the other direction and not even keeping the message from the wire. That way if the error is mutated (e.g. a different `Namespace` object is set, it is serialized correctly).

cretz: I wouldn't put too much thought into it, just keep the status you're given like other similar errors

bergundy: Lol, I'll just do that to move things along. I wanted to improve what we had, but I can see arguments in both directions.

bergundy: Done.

cretz (approved):

stephanos (approved):

serviceerror/namespace_unavailable.go
```diff
@@ -0,0 +1,81 @@
...
+	// NamespaceUnavailable is returned by the service when a request addresses a namespace that is unavailable. For
+	// example, when a namespace is in the process of failing over between clusters. This is a transient error that
+	// should be automatically retried by clients.
+	NamespaceUnavailable struct {
+		Namespace string
+		st        *status.Status
+	}
+)
+
+// NewNamespaceUnavailable returns new NamespaceInvalidState error.
```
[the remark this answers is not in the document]
bergundy: Thanks, I should have replaced with `/g`.

PR #196, opened 2024-11-25
Fix for protogen cmd

anekkanti:
**What changed?**
A couple of fixes for the `protogen` cmd.

- When looking for proto files ignore directories.
- Fix how regex is used when looking up enums.

**How did you test it?**

Ran `make proto` and verified that there was no diff generated.


yuandrew:
Closing this old PR, feel free to reopen if still working on it

PR #200, opened 2024-12-10
GHA check sdk-go compatibility

rodrigozhou:
**What changed?**
GHA check sdk-go compatibility

**Why?**
Automation: be able to check if a api-go version breaks sdk-go

**How did you test it?**

**Potential risks**

### DIFF
--- a/.github/workflows/check-sdk-compat.yml
+++ b/.github/workflows/check-sdk-compat.yml
@@ -32,19 +32,11 @@ jobs:
         id: inputs
         env:
           GH_TOKEN: ${{ github.token }}
-          SDK_REF: ${{ github.event.inputs.sdk_ref }}
-          API_REF: ${{ github.event.inputs.api_ref }}
+          SDK_REF: ${{ inputs.sdk_ref }}
+          API_REF: ${{ inputs.api_ref }}
         run: |
           if [[ "$SDK_REF" == "latest" ]]; then
             SDK_REF=$(gh api /repos/temporalio/sdk-go/releases/latest --jq '.name')
-          fi
-          if ! gh api "/repos/temporalio/sdk-go/commits/$SDK_REF"; then
-            echo "::error::Reference $SDK_REF not found in sdk-go"
-            exit 1
-          fi
-          if ! gh api "/repos/temporalio/api-go/commits/$API_REF"; then
-            echo "::error::Reference $API_REF not found in api-go"
-            exit 1
           fi
           echo "SDK_REF=$SDK_REF" >> "$GITHUB_OUTPUT"
           echo "API_REF=$API_REF" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/check-sdk-compat.yml
+++ b/.github/workflows/check-sdk-compat.yml
@@ -38,8 +38,8 @@ jobs:
           if [[ "$SDK_REF" == "latest" ]]; then
             SDK_REF=$(gh api /repos/temporalio/sdk-go/releases/latest --jq '.name')
           fi
-          echo "SDK_REF=$SDK_REF" >> "$GITHUB_OUTPUT"
-          echo "API_REF=$API_REF" >> "$GITHUB_OUTPUT"
+          echo "SDK_REF=$SDK_REF" | tee -a "$GITHUB_OUTPUT"
+          echo "API_REF=$API_REF" | tee -a "$GITHUB_OUTPUT"
 
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


cretz (approved):

.github/workflows/check-sdk-compat.yml
```diff
@@ -0,0 +1,84 @@
...
+        required: true
+        default: master
+
+jobs:
+  check-sdk-go-compatibility:
+    name: "Check sdk-go compatibility"
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Validate inputs
```
bergundy, line 31: Probably redundant since the checkout step will fail if the ref isn't valid.

.github/workflows/check-sdk-compat.yml
```diff
@@ -0,0 +1,84 @@
...
+        id: inputs
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SDK_REF: ${{ github.event.inputs.sdk_ref }}
+          API_REF: ${{ github.event.inputs.api_ref }}
+        run: |
+          if [[ "$SDK_REF" == "latest" ]]; then
+            SDK_REF=$(gh api /repos/temporalio/sdk-go/releases/latest --jq '.name')
```
bergundy, line 39: I see you're using the `GH_TOKEN` env var so probably not much of a concern but this command may get rate limited by GH.

I vote for removing this step to have one less moving part.

bergundy (commented):
Arguably this should have triggered a workflow in the Go SDK repo but this is fine.

.github/workflows/check-sdk-compat.yml
rodrigozhou: I need to at least translate the "latest" input to an actual tag. The `checkout` step fails with the "latest" ref.

bergundy:
This should run before a release is tagged and it's too late.

.github/workflows/check-sdk-compat.yml
rodrigozhou: The rate limit is 15000 requests per hour: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-github_token-in-github-actions

.github/workflows/check-sdk-compat.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: stable
+
+      - name: Update api-go
+        env:
+          API_REF: ${{ steps.inputs.outputs.API_REF }}
+        run: |
+          for f in $(find . -iname go.mod); do
```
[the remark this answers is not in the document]
rodrigozhou: It will error out.

cretz:
> Arguably this should have triggered a workflow in the Go SDK repo but this is fine.

I requested it be put in here so it can be updated by anyone using this repo that may need to change instead of just something owned by the SDK team, and it's a condition on this repo not the SDK one. Granted we plan on making sure the Go SDK updates to latest API release on each release it makes as an inverse condition.

.github/workflows/check-sdk-compat.yml
bergundy: Got it, yeah, I had it in my head that we want to verify master and not latest, but I think latest is right.
We need to give guidance though that sometimes `latest` is expected to fail, e.g. when making backwards incompatible changes, and that the action should be rerun with `master`. If that also fails, we should not cut a release.

bergundy (commented):
IMHO the overall process should be:
- Validate Go SDK
- Cut a draft api release
- Cut a draft api-go release
- Send slack notification for approval (can come later)
- Reviewer approves the releases (can say this needs both SDK and server)
- Release api
- Release api-go

cretz:
Works for me. I think the first 3 can be part of the same workflow or workflow triggers we have here and in other PRs.

.github/workflows/check-sdk-compat.yml
```diff
@@ -0,0 +1,76 @@
...
+        run: go run . check
+        working-directory: ./internal/cmd/build
+
+      - name: Run unit test
+        run: go run . unit-test
+        working-directory: ./internal/cmd/build
+
+      - name: Run integration tests
+        run: go run . integration-test -dev-server
+        working-directory: ./internal/cmd/build
```
bergundy, line 76: Integration tests have been flaky, I wonder if we can skip those. We mostly care that compilation isn't broken.

bergundy (approved):

.github/workflows/check-sdk-compat.yml
rodrigozhou: Added `continue-on-error: true` so even if the integration tests fails, it won't fail the job.

PR #205, opened 2025-01-23
Remove disabling UTF-8 validation hack

dnr:
**What changed?**
Remove the hack to allow disabling UTF-8 validation on our Go protos.

**Why?**
We should allow protobuf to validate all strings are valid UTF-8 as specified.

**How did you test it?**
Will be tested by server, sdk, and other components.

**Potential risks**
Anyone still using invalid UTF-8 will see their grpc requests fail. We've been monitoring for this in Temporal Cloud for most of a year now and there's no more invalid data. Self-hosted users should do the same validation before upgrading to a server or Go SDK release that contains this change (instructions [here](https://github.com/temporalio/temporal/releases/tag/v1.24.0)).


tdeebswihart (approved):
The diff stats are horrifying for this change. Thank you for following this saga through to its conclusion!

cretz (approved):
Echoing @tdeebswihart, thank you very much for sticking with it

bergundy (approved):

PR #209, opened 2025-02-10
Fix update.Response getting skipped by the proxy

Quinn-With-Two-Ns:
**What**

Add `update.Response` to proxy

**Why?**

Without this change payloads under `update.Response` are skipped by the proxy. This can cause update results to not go through the proxy.

**How did you test it?**

New unit test added


cretz (approved):
Thanks! Can we open a separate issue to have a test that somehow makes sure every payload is visited even if it's not reachable from workflow service? I can think of a few ways to write it.

Quinn-With-Two-Ns:
@cretz https://github.com/temporalio/sdk-go/issues/1809 done

stephanos (approved):

Quinn-With-Two-Ns:
@cretz Can you please re-review I also noticed we didn't handle any in the failure case

cretz (approved):

PR #210, opened 2025-02-14
Improve proxy codegen by using proto descriptors

Sushisource:
Closes https://github.com/temporalio/sdk-go/issues/1811

**What changed?**
Title. More reliable way of discovering everything

**Why?**
Avoid missing newly added top-level paths to payloads/failures

**How did you test it?**
Existing tests / comparing generated output.

**Potential risks**
Too much visiting!


proxy/interceptor.go
```diff
@@ -619,6 +620,58 @@ func visitPayloads(
 				return err
 			}
 
+		case *errordetails.MultiOperationExecutionFailure:
+
```
Sushisource, line 624: So a bunch of stuff got added to the failure interceptor, which previously I guess did not start from the top of anything that contains a failure, like the payload one does. Is that intentional? Assuming it is, is starting from the top like this a bad thing?

proxy/interceptor.go
```diff
@@ -2884,6 +2939,64 @@ func visitFailures(ctx *VisitFailuresContext, options *VisitFailuresOptions, obj
 				return err
 			}
 
+		case *command.ScheduleNexusOperationCommandAttributes:
```
Sushisource, line 2942: So a bunch of stuff got added to the failure interceptor, which previously I guess did not start from the top of anything that contains a failure, like the payload one does. Is that intentional? Assuming it is, is starting from the top like this a bad thing?

proxy/interceptor.go
Sushisource, line 624: Some things did, in fact, get missed!

proxy/interceptor.go
```diff
@@ -3041,6 +3157,18 @@ func visitFailures(ctx *VisitFailuresContext, options *VisitFailuresOptions, obj
 				return err
 			}
 
+		case *history.NexusOperationCompletedEventAttributes:
+			if o == nil {
+				continue
+			}
+			ctx.Parent = o
+			if err := visitFailures(
```
Quinn-With-Two-Ns, line 3165: This looks wrong, you are calling `visitFailures` with no objs? That is a no-op

proxy/interceptor.go
```diff
@@ -3298,6 +3450,19 @@ func visitFailures(ctx *VisitFailuresContext, options *VisitFailuresOptions, obj
...
 
+		case *workflow.WorkflowExecutionConfig:
+			if o == nil {
+				continue
+			}
+			ctx.Parent = o
+			if err := visitFailures(
+				ctx,
+				options,
+				o.GetUserMetadata(),
```
Quinn-With-Two-Ns, line 3461: `GetUserMetadata` doesn't contain a failure so why would we need to search it?

proxy/interceptor.go
Quinn-With-Two-Ns: I am not sure if the proxy ever look into error details though...

proxy/interceptor.go
```diff
@@ -2884,6 +2939,64 @@ func visitFailures(ctx *VisitFailuresContext, options *VisitFailuresOptions, obj
...
+			ctx.Parent = o
+			if err := visitFailures(
+				ctx,
+				options,
+				o.GetDetails(),
+			); err != nil {
+				return err
+			}
+
+		case *errordetails.QueryFailedFailure:
```
Quinn-With-Two-Ns, line 2987: I think this one we actually do need to handle because their is a payload here?

proxy/interceptor.go
Sushisource: What do you mean? It won't call with them? If so, oh well, no harm having it here anyway since they do contain payloads

proxy/interceptor.go
Sushisource: It looks like maybe some of these are still reporting as being needed to search because they have payloads rather than failures, I'll see if that's it.

Sushisource:
I changed the make stuff to generate the proto descriptors on the fly since checking in a 1mb file is annoying. However that would mean adding proto to CI which isn't there currently. Should we just add it?

proxy/interceptor.go
Quinn-With-Two-Ns: Yeah because these are part of the error

> It won't call with them? If so, oh well

Well the consequence is a payload or failure that should be encoded would not be so it is kinda important no?

Sushisource: Oh, I see what you mean now.

Well that's because it's not referenced by any other proto. In fact I don't understand this at all:

```
    // ExecuteMultiOperation executes multiple operations within a single workflow.
    //
    // Operations are started atomically, meaning if *any* operation fails to be started, none are,
    // and the request fails. Upon start, the API returns only when *all* operations have a response.
    //
    // Upon failure, it returns `MultiOperationExecutionFailure` where the status code
    // equals the status code of the *first* operation that failed to be started.
    //
    // NOTE: Experimental API.
    rpc ExecuteMultiOperation (ExecuteMultiOperationRequest) returns (ExecuteMultiOperationResponse) {
```

Quinn-With-Two-Ns: I think it is part of the error details https://grpc.io/docs/guides/error/#richer-error-model

Quinn-With-Two-Ns: The proxy doesn't look at errors at all right now AFAIK @cretz please correct me if I am wrong

Sushisource: Oh, I see, OK that makes sense. Yeah, I agree that is a possible gap. I also think probably another PR.

Sushisource:
Oops, didn't mean to have this still be draft. Need to resolve the CI question tho

proxy/interceptor.go
cretz: > The proxy doesn't look at errors at all right now AFAIK @cretz please correct me if I am wrong

Correct, ref https://github.com/temporalio/api-go/blob/dad8b169ada911d3778e070484d1ae78a58bd22b/proxy/interceptor.go#L99-L101. The only cases I am aware of where user-provided payloads can be in an error is `ExecuteMultiOperation` and `QueryWorkflow`. I agree it is a hole in our logic in the interceptor that we should fix (walk the gRPC status details). I will open an issue. But yes we should keep them in this switch statement of course.

internal/protojson/testprotos/enums/enums.proto
cretz, line 1: Hrmm, I didn't dig, but why are these files appearing now?

proxy/interceptor.go
cretz: I do not think it's a bad thing, on the contrary it is a good thing. I think the switch statements for payloads should work for every proto message that might contain a payload in any way. Similarly, I think the switch statements for failures should work for every proto message that might contain a failure in any way.

But @Quinn-With-Two-Ns's other comment alter is right, some of these seem to be calling `visitFailures` with no messages to visit.

proxy/interceptor.go
cretz: Yes, we need to handle that as part of `visitPayloads` (`visitFailures` here is a separate concern)

proxy/interceptor.go
Sushisource: Great, yeah that's what I thought too. I'll sort out the visit with nothing problem

internal/protojson/testprotos/enums/enums.proto
cretz: Hrmm, ok. I forget, but I guess as part of the the tests we added the generated proto code in we never committed the protos themselves?

Sushisource: Exactly

cmd/proxygenerator/interceptor.go
```diff
@@ -500,7 +674,14 @@ func walk(desired []types.Type, typ types.Type, records *map[string]*TypeRecord,
 		// All the Get... methods return the relevant protobuf as the first result
 		resultType := sig.Results().At(0).Type()
 
-		if checkDirectPayload && resultType.String() == "*go.temporal.io/api/common/v1.Payload" {
+		hasDirectMatch := false
+		for _, directMatchType := range directMatchTypes {
+			if resultType.String() == types.NewPointer(directMatchType).String() {
+				hasDirectMatch = true
```
Sushisource, line 680: Don't love that I had to do this but I could not for the life of me get reflection to be equal properly

proxy/interceptor.go
Sushisource: Ok these were fixed by removing the hardcoded check for payloads in `walk`

PR #213, opened 2025-03-24
Support proxy interceptors visiting Payloads and Failures in gRPC errors

yuandrew:
**What changed?**
Added support for proxy interceptors visiting gRPC errors. 

**Why?**
In some scenarios there can be user-provided payloads.
https://github.com/temporalio/sdk-go/issues/1825

**How did you test it?**
Added new test to validate both these behaviors

**Potential risks**
Proxy interceptors could completely break?


cmd/proxygenerator/interceptor.go
```diff
@@ -104,7 +105,22 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+					return fmt.Errorf("Error returned from rpc invoker should be a status.Status")
+				}
+				// user provided payloads can sometimes end up in the status details of
+				// gRPC errors, make sure to visit those as well
+				for _, detail := range stat.Details() {
+					detailAny, ok := detail.(*anypb.Any)
+					if !ok {
+						return fmt.Errorf("Error returned from rpc invoker should be anypb.Any")
+					}
+					VisitPayloads(ctx, detailAny, *options.Inbound)
```
cretz, line 120: So this is a bit expensive to do for every error ever when we know that it only happens via query I think (and maybe multi-op in the future). I wonder if we should narrow this down. For instance, maybe we should only perform this visit if the any's type URL contains `temporal.api.errordetails.v1` or even better, only if it contains the fully qualified name of a message in that proto package that we have programmatically determined has payloads beneath it (so only `temporal.api.errordetails.v1.QueryFailedFailure` and `temporal.api.errordetails.v1.MultiOperationExecutionFailure`).

proxy/interceptor.go
cretz, line 1: So there's an issue here with failure converter interceptor and payload codec interceptor order (that kinda doesn't to this PR). If you set the failure converter to use encoded failures and you have the failures interceptor _after_ the payload visitor interceptor, you may miss intercepting payloads _created_ by the failure converter.

I think we should consider documenting above `NewFailureVisitorInterceptor` and `NewPayloadVisitorInterceptor` in this package, and above `NewFailureGRPCClientInterceptor` and `NewPayloadCodecGRPCClientInterceptor` in the `go.temporal.io/sdk/converter` package that failure converter should come before payload codec converter. But this can be done in a separate issue, just now thinking about it.

cmd/proxygenerator/interceptor.go
```diff
@@ -104,7 +105,22 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
 
 		err := invoker(ctx, method, req, response, cc, opts...)
 		if err != nil {
-			return err
+			if options.Inbound != nil {
+				stat, ok := status.FromError(err)
+				if !ok {
+					return fmt.Errorf("Error returned from rpc invoker should be a status.Status")
```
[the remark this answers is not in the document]
yuandrew: got it, could you explain why not wrapping the err is being more cautious? My mental model is richer error context is more cautious/better.

cmd/proxygenerator/interceptor.go
yuandrew: good suggestion! Left a TODO for https://github.com/temporalio/sdk-go/issues/1864, as we're unable to parse MultiOperationExecutionFailure due to not supporting `repeated Any` fields yet

cmd/proxygenerator/interceptor.go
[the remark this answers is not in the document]
cretz: IMO we don't want our inability to convert this error to change the user error or error expectations they may have. For instance, if the user happens to already have converted the error to something of theirs in an inner interceptor, this will change what they did (not that they should be doing that, but we should allow it).

cmd/proxygenerator/interceptor.go
cretz: My only concern hardcoding this is that if we add a new `errordetails` in the future we'll forget to come back and update this. I wonder if we can programmatically determine at code generation time the set of qualified message names that have an "any" or "proto" anywhere in their hierarchy? If it's too much work we can discuss alternatives.

yuandrew: hmm, I can give that a shot, how would we cache the results of this check? Is it okay to cache it globally? I'm imaging a function that produces a list of types that contain `any` or `failure` (did you mean `failure` instead of `proto`?), but not seeing a parent struct or anywhere we can cache the results

cretz: > how would we cache the results of this check?

No cache is needed. What I am suggesting is to derive the list at code gen time. Can still code gen a bunch of `==` like you have now, I'm just saying don't hardcode them in the code generator, derive them. Having them hardcoded in the final output is still ok/good.

cmd/proxygenerator/interceptor.go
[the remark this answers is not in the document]
yuandrew: got it, thx for the explanation

proxy/interceptor.go
yuandrew: https://github.com/temporalio/sdk-go/issues/1886

cretz (approved):

proxy/interceptor.go
```diff
@@ -98,7 +99,24 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
 
 		err := invoker(ctx, method, req, response, cc, opts...)
 		if err != nil {
-			return err
+			if options.Inbound != nil {
+				stat, ok := status.FromError(err)
+				if !ok {
+					return err
+				}
```
cretz, line 106: Now that I think about it, I am a tad concerned about this logic because I am concerned that both `response` _and_ non-status `err` can be non-nil, but since technically an interceptor can return an error. I think we may still want to apply the logic below this `err != nil` block even on error.

proxy/interceptor.go
```diff
@@ -98,7 +99,24 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+				// user provided payloads can sometimes end up in the status details of
+				// gRPC errors, make sure to visit those as well
+				for _, detail := range stat.Details() {
+					detailAny, ok := detail.(*anypb.Any)
+					if detailAny.MessageName() == "temporal.api.errordetails.v1.QueryFailedFailure" || detailAny.MessageName() == "temporal.api.errordetails.v1.MultiOperationExecutionFailure" {
+						if !ok {
+							return fmt.Errorf("Error returned from rpc invoker should be anypb.Any")
+						}
```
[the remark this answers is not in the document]
yuandrew: hmm, I think this is still needed, just in a different spot, if detailAny is nil pointer, wouldn't calling `MessageName()` cause a runtime panic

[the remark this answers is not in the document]
cretz: No I believe it is nil safe (which is why it didn't cause a panic when you had this check _after_ you did the `MessageName()` call above it in the `if`). But I am ok with the check anyways, but like in other situations, I would not recommend erroring, just don't visit if not `ok`. We only want to affect gRPC error and response if they have payload, otherwise we don't want to do anything to any value that may occur in the interceptor.

proxy/interceptor.go
```diff
@@ -97,8 +98,22 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+				for _, detail := range stat.Details() {
+					detailAny, ok := detail.(*anypb.Any)
+					if !ok {
+						return fmt.Errorf("Error returned from rpc invoker should be anypb.Any")
+					}
+					if detailAny.MessageName() == "temporal.api.errordetails.v1.QueryFailedFailure" || detailAny.MessageName() == "temporal.api.errordetails.v1.MultiOperationExecutionFailure" {
+						VisitPayloads(ctx, detailAny, *options.Inbound)
+					}
+				}
+				return err
```
cretz, line 115: Do not `return err` here, let this fall through to the existing behavior and change the final `return nil` to `return err`

proxy/interceptor.go
```diff
@@ -97,8 +98,22 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+				// user provided payloads can sometimes end up in the status details of
+				// gRPC errors, make sure to visit those as well
+				for _, detail := range stat.Details() {
+					detailAny, ok := detail.(*anypb.Any)
+					if !ok {
+						return fmt.Errorf("Error returned from rpc invoker should be anypb.Any")
+					}
+					if detailAny.MessageName() == "temporal.api.errordetails.v1.QueryFailedFailure" || detailAny.MessageName() == "temporal.api.errordetails.v1.MultiOperationExecutionFailure" {
+						VisitPayloads(ctx, detailAny, *options.Inbound)
+					}
```
cretz, line 113:
```suggestion
					if ok && (detailAny.MessageName() == "temporal.api.errordetails.v1.QueryFailedFailure" || detailAny.MessageName() == "temporal.api.errordetails.v1.MultiOperationExecutionFailure") {
						VisitPayloads(ctx, detailAny, *options.Inbound)
					}
```

Do not return our own custom errors here if the types do not conform

proxy/interceptor.go
yuandrew: This changes existing behavior today, I think it's better if I tackle this in a separate PR with dedicated tests to validate this scenario.

proxy/interceptor_test.go
```diff
@@ -399,6 +401,66 @@ func TestClientInterceptor(t *testing.T) {
...
+	_, err = client.ExecuteMultiOperation(context.Background(), &workflowservice.ExecuteMultiOperationRequest{})
+	// We expect that even though an error is returned, the Payload visitor visited the payload
+	// included in the GRPC error details
+	require.Error(err)
+	// TODO: https://github.com/temporalio/sdk-go/issues/1864
+	// This check should be switched to True once the issue above is closed
+	require.False(proto.Equal(inputs.Payloads[0], inboundPayload))
+
+	_, err = client.QueryWorkflow(context.Background(), &workflowservice.QueryWorkflowRequest{})
+	require.Error(err)
```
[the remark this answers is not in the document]
yuandrew: ahh thank you for pointing this out.. I completely missed it. Just pushed a fix

cmd/proxygenerator/interceptor.go
```diff
@@ -104,6 +105,28 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+					for _, detail := range stat.Details() {
+						detailAny, ok := detail.(*anypb.Any)
+						if ok && {{ range $i, $name := .GrpcPayload }}{{ if $i }} || {{ end }}detailAny.MessageName() == "{{$name}}"{{ end }} {
+							err = VisitPayloads(ctx, detailAny, *options.Inbound)
+							if err != nil {
+								return err
+							}
+						}
+						newStatus, err = newStatus.WithDetails(detailAny)
+						if err != nil {
```
yuandrew, line 123: @cretz just want to confirm that returning an error here is fine, pretty confident this is different than the previously mentioned scenario of returning custom errors, but doesn't hurt to ask 🤷‍♂️

proxy/interceptor.go
cretz: Hrmm, I see we did always eagerly return any error that happened and not affect response. Agreed can be done separately.

proxy/interceptor.go
```diff
@@ -98,6 +99,28 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+					newStatus := status.New(stat.Code(), stat.Message())
+					for _, detail := range stat.Details() {
+						detailAny, ok := detail.(*anypb.Any)
+						if ok && detailAny.MessageName() == "temporal.api.errordetails.v1.QueryFailedFailure" || detailAny.MessageName() == "temporal.api.errordetails.v1.MultiOperationExecutionFailure" {
+							err = VisitPayloads(ctx, detailAny, *options.Inbound)
+							if err != nil {
+								return err
+							}
+						}
+						newStatus, err = newStatus.WithDetails(detailAny)
```
cretz, line 116: Sorry to be pedantic, but can we collect the details into a slice and just call `WithDetails` once after the loop?

proxy/interceptor.go
```diff
@@ -98,6 +99,28 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
 		if err != nil {
+			if options.Inbound != nil {
+				stat, ok := status.FromError(err)
+				if ok {
+					// user provided payloads can sometimes end up in the status details of
+					// gRPC errors, make sure to visit those as well
+					newStatus := status.New(stat.Code(), stat.Message())
+					for _, detail := range stat.Details() {
+						detailAny, ok := detail.(*anypb.Any)
+						if ok && detailAny.MessageName() == "temporal.api.errordetails.v1.QueryFailedFailure" || detailAny.MessageName() == "temporal.api.errordetails.v1.MultiOperationExecutionFailure" {
```
cretz, line 110:
```suggestion
						if ok && (detailAny.MessageName() == "temporal.api.errordetails.v1.QueryFailedFailure" || detailAny.MessageName() == "temporal.api.errordetails.v1.MultiOperationExecutionFailure") {
```

Can we add parens when changing the comparison operator here?

cmd/proxygenerator/interceptor.go
cretz: I think we're ok returning err here like you do

proxy/interceptor.go
yuandrew: I tried that originally, but ran into `Cannot use 'detailAnys' (type []*anypb. Any) as the type [] protoadapt. MessageV1`, couldn't figure out how to get the slice to work

cretz: Make a slice of `[]any` and append to that.

cretz (approved):

proxy/interceptor.go
yuandrew: https://github.com/temporalio/sdk-go/issues/1889

cmd/proxygenerator/interceptor.go
```diff
@@ -104,6 +109,30 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
 		if err != nil {
+			if options.Inbound != nil {
+				stat, ok := status.FromError(err)
+				if ok {
+					// user provided payloads can sometimes end up in the status details of
+					// gRPC errors, make sure to visit those as well
+					newStatus := status.New(stat.Code(), stat.Message())
+					var newDetails []protoadapt.MessageV1
+					for _, detail := range stat.Details() {
+						detailAny, ok := detail.(*anypb.Any)
```
Quinn-With-Two-Ns, line 120: What if details contains a proto message that is not any but contains a payload?

proxy/interceptor_test.go
```diff
@@ -478,6 +571,54 @@ func (t *testGRPCServer) PollActivityTaskQueue(
 	}, nil
 }
 
+func (t *testGRPCServer) ExecuteMultiOperation(
```
Quinn-With-Two-Ns, line 574: Did we verify this matches the real output of the server?

cmd/proxygenerator/interceptor.go
cretz: In theory this should be impossible looking at source, but I agree, we might as well first type assert to message v1, _then_ optionally type assert to any (falling back to just appending the v1 message if it didn't work)

proxy/interceptor_test.go
Quinn-With-Two-Ns: multi op is supported, it is used by update with start

cmd/proxygenerator/interceptor.go
Quinn-With-Two-Ns: by "theory this should be impossible looking at source" do you mean the server source or go protobuf?

PR #214, opened 2025-03-27
Support parsing error details that aren't anypb.Any type

yuandrew:
**What changed?**
Add support for parsing payloads for non-Any details. 

Also added in support for visiting Payload/failures when both response and non-status errors are non-nil https://github.com/temporalio/sdk-go/issues/1889

**Why?**
This shouldn't be possible, but adding a safety net here.

Follow up to https://github.com/temporalio/api-go/pull/213, 

**How did you test it?**

**Potential risks**


cmd/proxygenerator/interceptor.go
```diff
@@ -109,39 +109,65 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+					newStatus, err = newStatus.WithDetails(detailAny)
 					if err != nil {
 						return err
 					}
-					return newStatus.Err()
+				} else if okDetail {
+					msg := protoadapt.MessageV2Of(detailV1)
+					if slices.Contains(payloadTypes, string(msg.ProtoReflect().Descriptor().FullName())) {
+						err = VisitPayloads(ctx, msg, *inbound)
+					}
```
cretz, line 151: I think you can remove this part, I don't think we need to visit non-Any details (which should never happen

cmd/proxygenerator/interceptor.go
```diff
@@ -109,39 +109,65 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+		stat, ok := status.FromError(err)
+		if ok {
+			// user provided payloads can sometimes end up in the status details of
+			// gRPC errors, make sure to visit those as well
+			newStatus := status.New(stat.Code(), stat.Message())
+			var newDetails []protoadapt.MessageV1
+			for _, detail := range stat.Details() {
+				detailAny, okAny := detail.(*anypb.Any)
+				detailV1, okDetail := detail.(protoadapt.MessageV1)
+				payloadTypes := []string{ {{ range $i, $name := .GrpcPayload }}{{ if $i }}, {{ end }}"{{$name}}"{{ end }} }
```
cretz, line 134: Do this outside of the function as a private package level var

cmd/proxygenerator/interceptor.go
```diff
@@ -109,39 +109,65 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+	}, nil
+}
+
+func parseGrpcPayload(ctx context.Context, err error, inbound *VisitPayloadsOptions) error {
+	if inbound != nil {
+		stat, ok := status.FromError(err)
+		if ok {
+			// user provided payloads can sometimes end up in the status details of
+			// gRPC errors, make sure to visit those as well
+			newStatus := status.New(stat.Code(), stat.Message())
```
cretz, line 129: You should not do any of the logic in this function if there are no details IMO

cmd/proxygenerator/interceptor.go
```diff
@@ -109,39 +109,65 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+		if resMsg, ok := response.(proto.Message); ok && options.Inbound != nil {
+			return VisitPayloads(ctx, resMsg, *options.Inbound)	
+		}
+		
+		return nil
+	}, nil
+}
+
+func parseGrpcPayload(ctx context.Context, err error, inbound *VisitPayloadsOptions) error {
+	if inbound != nil {
```
cretz, line 124: You can short-circuit (e.g. `if inbound == nil { return err }`) if that's easier for you

cmd/proxygenerator/interceptor.go
```diff
@@ -109,39 +109,65 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+		stat, ok := status.FromError(err)
+		if ok {
+			// user provided payloads can sometimes end up in the status details of
+			// gRPC errors, make sure to visit those as well
+			newStatus := status.New(stat.Code(), stat.Message())
+			var newDetails []protoadapt.MessageV1
+			for _, detail := range stat.Details() {
+				detailAny, okAny := detail.(*anypb.Any)
```
cretz, line 132: This can be done inline in the if, e.g. `if msg, ok := detail.(*anypb.Any); ok {` if it's easier for you

cmd/proxygenerator/interceptor.go
```diff
@@ -109,39 +109,65 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+						err = VisitPayloads(ctx, msg, *inbound)
+					}
+					newDetails = append(newDetails, detailV1)
+				} else {
+					// stat.Details() returned an error
+					detailErr, ok := detail.(error)
+					if !ok {
+						return fmt.Errorf("unreachable")
+					}
+						return detailErr
```
cretz, line 159: Can probably remove this, I'm ok if we just don't append details that don't match message v1 or anypb.Any (they will always be anypb.Any anyways looking at code).

Actually, looking back at this, if I were you, I'd operate on the raw status proto. So my function may look like:

```go
func parseGrpcPayload(ctx context.Context, err error, inbound *VisitPayloadsOptions) error {
  if err == nil || inbound == nil {
    return err
  }
  s, ok := status.FromError(err)
  if !ok {
    return err
  }
  p := s.Proto()
  for _, detail := range p.Details {
    if slices.Contains(payloadTypes, string(detailAny.MessageName())) {
      if vErr := VisitPayloads(ctx, detail, *inbound); vErr != nil {
        return vErr
      }
    }
  }
  return status.ErrorProto(p)
}
```

cmd/proxygenerator/interceptor.go
yuandrew: This was @Quinn-With-Two-Ns's original main concern, if one day we _do_ decide to have a non-Any detail. Quinn, correct me if I misunderstood anything

yuandrew:
oops, I kinda forgot about this PR 😅

cmd/proxygenerator/interceptor.go
```diff
@@ -112,31 +112,11 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
-						}
-						newDetails = append(newDetails, detailAny)
-					}
-					newStatus, err = newStatus.WithDetails(newDetails...)
-					if err != nil {
-						return err
-					}
-					return newStatus.Err()
-				}
+			if s, ok := status.FromError(err); ok {
```
cretz, line 115: Don't have to convert to status if there is no inbound, might as well do `status.FromError` inside of the `parseGrpcPayload` _after_ you've checked whether there even is an inbound

cmd/proxygenerator/interceptor.go
```diff
@@ -147,6 +127,25 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
 	}, nil
 }
 
+func parseGrpcPayload(ctx context.Context, err error, s *status.Status, inbound *VisitPayloadsOptions) error {
```
cretz, line 130: Pedantic, but this name is a little generic for something only for errors. Maybe `visitGrpcError` or something?

cmd/proxygenerator/interceptor.go
```diff
@@ -147,6 +127,25 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+	if inbound == nil {
+		return err
+	}
+	p := s.Proto()
+	for _, detail := range p.Details {
+		payloadTypes := []string{ {{ range $i, $name := .GrpcPayload }}{{ if $i }}, {{ end }}"{{$name}}"{{ end }} }
+		for _, payloadType := range payloadTypes {
+			if strings.Contains(detail.String(), payloadType) {
```
cretz, line 138: I don't think this should do a strings contains on the https://pkg.go.dev/google.golang.org/protobuf/types/known/anypb#Any.String call, rather you should check for message name

cmd/proxygenerator/interceptor.go
```diff
@@ -147,6 +127,25 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
+				if vErr := VisitPayloads(ctx, detail, *inbound); vErr != nil {
+					return vErr
+				}
+			}
+			break
```
cretz, line 143: This breaks after checking only the first payload type?

cmd/proxygenerator/interceptor.go
yuandrew: `detail.MessageName()` is `google.protobuf.Any` here. And after testing this just now, I found out this is a double wrapped any, so I would have to unmarshal twice to get to the underlying type. 

It should be more efficient to do the current `detail.String()` contains check, than to unmarshal each any twice, to then perform this check

cmd/proxygenerator/interceptor.go
yuandrew: good catch, put it before the wrong bracket 😅

cmd/proxygenerator/interceptor.go
yuandrew: I moved the status check out to allow the scenario for `response` and non-status `err` to also visit `response`. What you say makes sense, will move the inbound check out to before the `status.FromError` check outside the function

cmd/proxygenerator/interceptor.go
yuandrew: ahh yes, my test was unnecessarily adding an extra `Any` layer https://github.com/temporalio/api-go/blob/master/proxy/interceptor_test.go#L648

cmd/proxygenerator/interceptor.go
```diff
@@ -112,31 +112,14 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
-				}
+			if options.Inbound == nil {
+				return err
+			}
+			if s, ok := status.FromError(err); ok {
+				// user provided payloads can sometimes end up in the status details of
+				// gRPC errors, make sure to visit those as well
+				return visitGrpcErrorPayload(ctx, err, s, options.Inbound)
 			}
-			return err
```
cretz, line 139: I think this `return err` is still needed here, I don't think you want to fall through to success stuff if `status.FromError` happens to not be `ok`.

cmd/proxygenerator/interceptor.go
```diff
@@ -147,6 +130,19 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
 	}, nil
 }
 
+func visitGrpcErrorPayload(ctx context.Context, err error, s *status.Status, inbound *VisitPayloadsOptions) error {
+	p := s.Proto()
+	for _, detail := range p.Details {
+		payloadTypes := []string{ {{ range $i, $name := .GrpcPayload }}{{ if $i }}, {{ end }}"{{$name}}"{{ end }} }
```
cretz, line 136: I'd move this to a package-level var (same for `failureTypes` below), no need to make this fixed, immutable slice in this function (much less every loop iteration)

yuandrew: shoot, I thought I did this before from your previous comment, thanks for the reminder

cmd/proxygenerator/interceptor.go
yuandrew: Maybe I'm misunderstanding something, https://github.com/temporalio/api-go/pull/213#discussion_r2014094412, I thought the original desire was to not immediately return err here.

Oh but I guess if a `response` and `err` are both non-nil, we want to visit the payload, but we'd still want to return the error, right?

yuandrew: But if both `response` and `err` are non-nil, would we return the err, or `return VisitPayloads(ctx, resMsg, *options.Inbound)` from processing the response?

cmd/proxygenerator/interceptor.go
```diff
@@ -111,42 +114,34 @@ func NewPayloadVisitorInterceptor(options PayloadVisitorInterceptorOptions) (grp
...
-					if err != nil {
-						return err
-					}
-					return newStatus.Err()
-				}
+		if err != nil && options.Inbound != nil {
+			if s, ok := status.FromError(err); ok {
+				// user provided payloads can sometimes end up in the status details of
+				// gRPC errors, make sure to visit those as well
+				return visitGrpcErrorPayload(ctx, err, s, options.Inbound)
```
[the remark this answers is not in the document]
yuandrew: sorry! completely missed that from the suggestion

cretz (approved):

PR #218, opened 2025-04-22
💥[breaking] Remove cloud files

yuandrew:
**What changed?**
Removed all cloud files, now that they live at https://github.com/temporalio/cloud-sdk-go

**Why?**
Cloud files have been moved to their own repo, no need to keep them here anymore.

The Go SDK's has not used cloud protos for a few releases now, so this should be safe.

https://github.com/temporalio/sdk-go/issues/1921

**How did you test it?**

**Potential risks**


cretz (commented):
This is great, thanks! @anekkanti - can you confirm this looks good to you? @Quinn-With-Two-Ns - any concerns about newer API versions no longer working with older Go SDKs that include the cloud ops API? @yuandrew - can you put a 💥 on the PR title and clarify that it's a breaking change to an experimental thing, that way they know to call it out in release notes.

Quinn-With-Two-Ns:
As long as the most recent released Go SDK works I have no concern. @yuandrew can you just confirm that?

cretz (approved):
(confirmed off-PR that @anekkanti approves and that this package is not in use in latest released Go SDK)

PR #219, opened 2025-04-23
Re-enable test visiting payloads in error details

yuandrew:
**What changed?**

This TODO is stale, re-enabling the test. note that because of https://github.com/temporalio/sdk-go/issues/1862, we cannot visit an Any Payload, not sure if that's a scenario we need to support today

**Why?**
More test coverage!

**How did you test it?**

**Potential risks**


proxy/interceptor_test.go
```diff
@@ -498,9 +498,7 @@ func TestClientInterceptorGrpcFailures(t *testing.T) {
 	// We expect that even though an error is returned, the Payload visitor visited the payload
 	// included in the GRPC error details
 	require.Error(err)
-	// TODO: https://github.com/temporalio/sdk-go/issues/1864
-	// This check should be switched to True once the issue above is closed
-	require.False(proto.Equal(inputs.Payloads[0], inboundPayload))
+	require.True(proto.Equal(inputs.Payloads[0], inboundPayload))
```
cretz, line 501: Curious, so is this test failing in `main` today?

cretz (approved):

PR #221, opened 2025-05-12
Add New...f constructors for all service errors

alexshtin:
**What changed?**
Add `New...f` constructors for all service errors.

**Why?**
We have similar helpers in different places but they all belong to this repo.

**How did you test it?**
Didn't test.


serviceerror/already_exists.go
```diff
@@ -20,6 +21,13 @@ func NewAlreadyExist(message string) error {
 	}
 }
 
+// NewAlreadyExistf returns new AlreadyExists error with formatted message.
+func NewAlreadyExistf(format string, args ...interface{}) error {
```
dnr, line 25:
```suggestion
// NewAlreadyExistsf returns new AlreadyExists error with formatted message.
func NewAlreadyExistsf(format string, args ...interface{}) error {
```
and ideally the non-f one should be renamed too

serviceerror/multi_op.go
```diff
@@ -20,6 +21,11 @@ func NewMultiOperationExecution(message string, errs []error) error {
 	return &MultiOperationExecution{Message: message, errs: errs}
 }
 
+// NewMultiOperationExecutionf returns a new MultiOperationExecution error with formatted message.
+func NewMultiOperationExecutionf(format string, errs []error, args ...interface{}) error {
```
dnr, line 25: weird to have format and args separated, how about switching the order?

serviceerror/namespace_invalid_state.go
```diff
@@ -40,6 +40,16 @@ func NewNamespaceInvalidState(namespace string, state enumspb.NamespaceState, al
 	}
 }
 
+// NewNamespaceInvalidStatef returns new NamespaceInvalidState error with formatted message.
+func NewNamespaceInvalidStatef(namespace string, state enumspb.NamespaceState, allowedStates []enumspb.NamespaceState, format string, args ...interface{}) error {
```
dnr, line 44: do we really need this one? the main constructor does a good job with messages...

serviceerror/namespace_not_active.go
```diff
@@ -35,6 +35,16 @@ func NewNamespaceNotActive(namespace, currentCluster, activeCluster string) erro
 	}
 }
 
+// NewNamespaceNotActivef returns new NamespaceNotActive error with formatted message.
+func NewNamespaceNotActivef(namespace, currentCluster, activeCluster, format string, args ...interface{}) error {
```
dnr, line 39: same here...

serviceerror/namespace_unavailable.go
```diff
@@ -26,6 +26,14 @@ func NewNamespaceUnavailable(namespace string) error {
 	}
 }
 
+// NewNamespaceUnavailablef returns new NamespaceUnavailable error with formatted message.
+func NewNamespaceUnavailablef(namespace, format string, args ...interface{}) error {
```
dnr, line 30: this probably shouldn't exist, NamespaceUnavailable has no Message field

serviceerror/newer_build_exists.go
```diff
@@ -27,6 +27,14 @@ func NewNewerBuildExists(defaultBuildID string) error {
 	}
 }
 
+// NewNewerBuildExistsf returns new NewerBuildExists error with formatted message.
+func NewNewerBuildExistsf(defaultBuildID, format string, args ...interface{}) error {
```
dnr, line 31: maybe not needed?

serviceerror/resource_exhausted.go
```diff
@@ -26,6 +27,14 @@ func NewResourceExhausted(cause enumspb.ResourceExhaustedCause, message string)
 	}
 }
 
+// NewResourceExhaustedf returns new ResourceExhausted error with formatted message.
+func NewResourceExhaustedf(cause enumspb.ResourceExhaustedCause, format string, args ...interface{}) error {
```
dnr, line 31: should we have constructors that take Scope also? we always set it everywhere now, right? I feel like we'll never get to use this one because we need a Scope

serviceerror/server_version_not_supported.go
```diff
@@ -28,6 +28,15 @@ func NewServerVersionNotSupported(serverVersion, supportedVersions string) error
 	}
 }
 
+// NewServerVersionNotSupportedf returns new ServerVersionNotSupported error with formatted message.
+func NewServerVersionNotSupportedf(serverVersion, supportedVersions, format string, args ...interface{}) error {
```
dnr, line 32: feel like the existing message should be fine everywhere?

serviceerror/system_workflow.go
```diff
@@ -27,6 +27,14 @@ func NewSystemWorkflow(workflowExecution *common.WorkflowExecution, workflowErro
 	}
 }
 
+// NewSystemWorkflowf returns new SystemWorkflow error with formatted workflow error.
+func NewSystemWorkflowf(workflowExecution *common.WorkflowExecution, format string, args ...interface{}) error {
```
dnr, line 31: I can't imagine ever using this, it should always be constructed from an `error`

serviceerror/already_exists.go
```diff
@@ -13,13 +15,20 @@ type (
 	}
 )
 
-// NewAlreadyExist returns new AlreadyExists error.
-func NewAlreadyExist(message string) error {
```
cretz, line 17: We can't make backwards-incompatible changes to this package IMO (even if users shouldn't really be using it)

serviceerror/already_exists.go
```diff
@@ -13,13 +15,20 @@ type (
...
+// NewAlreadyExists returns new AlreadyExists error.
+func NewAlreadyExists(message string) error {
 	return &AlreadyExists{
 		Message: message,
 	}
 }
 
+// NewAlreadyExistsf returns new AlreadyExists error with formatted message.
+func NewAlreadyExistsf(format string, args ...any) error {
```
cretz, line 26: Users are not really expected to create these errors, do we really need new user-facing helpers? What is the use case driving this PR?

serviceerror/already_exists.go
alexshtin: It is not used by SDK. Only by server. But I agree... I should leave old constructor as deprecated.

serviceerror/already_exists.go
alexshtin: These errors are constructed mostly on server. User is a server dev. We have hundreds of usages with `fmt.Sprintf` and even small helpers like [this](https://github.com/temporalio/temporal/blob/3dcdfde5ac76bfb7922aa0e9d749dab7c105cfe4/service/history/workflow/update/util.go#L31). I want to get rid of all these helpers.

cretz: I don't think we need to update this user-facing package with server-dev-only utilities,. You aren't getting rid of the helpers, you're moving them to the user-facing stable library. Can you move them to somewhere specific to server repo?

serviceerror/already_exists.go
cretz: We do not know everyone that uses this. This is one of the most included libraries by Temporal users. Many users have proxies, client mocks, etc that may take advantage of these things. We should not adjust this library as if it's only used by server, github.com/temporalio/temporal is a better place for code that is only used by server.

serviceerror/already_exists.go
dnr: Go SDK has several instances of `serviceerror.New<Something>(fmt.Sprintf(...))`, it can and should also use these new constructors.

Also, it would be pretty weird to have the `NewSomething` constructor be in this package but the `NewSomethingf` constructor be in a whole other package.

dnr (approved):

serviceerror/already_exists.go
alexshtin: `serviceerror` package should be in this repo at all. This is a mistake I made long time ago. Having constructor separate from struct definition doesn't look good to me. And yes, if other users use `New<Something>` they will definitely benefit from these new constructors too.

cretz: > serviceerror package should NOT be in this repo at all

But it's used in the SDK and by users, what repo should it be in?

> Go SDK has several instances of serviceerror.New<Something>(fmt.Sprintf(...)), it can and should also use these new constructors.

Right, I am not questioning the validity of the methods, I am questioning the user-facing aspect. These types of things are added in the Go SDK because we expect users to use them and therefore we are ok with the increased API surface area.

My primary concern is adding server-only helpers to this repository. It is less of a concern with these of course since they are simple constructor overloads. But in general it's a bad pattern to add "server-developer-only helpers" to this user-facing library.

If y'all feel strongly enough about giving all users these helpers, ok, but we would prefer server-only code to be server-only code.

cretz: > They are not server-only

This is at odds with some of the other comments, e.g. "It is not used by SDK. Only by server", hence the concern/thread here. Marking approved.

cretz (approved):

serviceerror/already_exists.go
dnr: Not used by SDK _yet_ :slightly_smiling_face:

PR #223, opened 2025-06-20
Add "Aborted" servicerror

stephanos:
**What changed?**

Added `Aborted` serviceerror.

**Why?**

There's a need to return an `Aborted` serviceerror from the server.

Given that we these serviceerror types for each status code, it seems like the right thing to add this here.

**How did you test it?**

**Potential risks**

### DIFF
--- a/serviceerror/convert.go
+++ b/serviceerror/convert.go
@@ -94,7 +94,7 @@ func FromStatus(st *status.Status) error {
 		case *failure.MultiOperationExecutionAborted:
 			return newMultiOperationAborted(st)
 		default:
-			// fall through to st.Err()
+			return newAborted(st)
 		}
 	case codes.Internal:
 		switch errDetails := errDetails.(type) {
--- a/serviceerror/aborted.go
+++ b/serviceerror/aborted.go
@@ -1,0 +1,49 @@
+package serviceerror
+
+import (
+	"fmt"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type (
+	// Aborted represents an aborted error.
+	Aborted struct {
+		Message string
+		st      *status.Status
+	}
+)
+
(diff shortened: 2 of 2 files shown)


serviceerror/convert.go
```diff
@@ -94,7 +94,7 @@ func FromStatus(st *status.Status) error {
 		case *failure.MultiOperationExecutionAborted:
 			return newMultiOperationAborted(st)
 		default:
-			// fall through to st.Err()
+			return newAborted(st)
```
ychebotarev, line 97: not an issue.
It looks like the only different between those errors (like `return newUnavailable(st)`) is the name of the structure, and the rest is the same.
Any specific reason we choose this approach instead of "newErrorWithCode" or something like that, that looks like

```
ErrorWithCode struct {
		Message string
                 Code  codes.Code
		st      *status.Status
	}
``` 

and then other "concrete" errors inherits from it? This way we may avoid reflection.

ychebotarev (approved):

serviceerror/convert.go
cretz, line 97: I am (only slightly) concerned this may be technically incompatible behavior changing an error type that users may be using. Can we make sure that this is _clearly_ called out in the release notes of the next release of this repo?

cretz, line 97: What about out of range and unauthenticated? Should we do those now too or wait?

stephanos: Right; I also thought about that for a bit.

From what I understand, the purpose of `FromStatus` is to convert `status.Status` to a particular type of service error; or fallback to `status.Error` if there isn't one available.

An incompatibility issue would occur if a user 
(1) relied on the error message details (`.Error()`) or
(2) tried to cast the error to a `status.Error`

Both seem unlikely, but definitely technically possible.

👍 I can add a warning to the release notes.

PR #224, opened 2025-06-25
fix: add timestamp to trigger immediately signal

spkane31:
adding a timestamp to `TriggerImmediatelyRequest` message

We can make the wft “more idempotent”: put the desired trigger time in the trigger immediately request itself, so it goes in the signal payload, so all wfts attempt to start the target with the same id.

Covered by existing functional tests

**Potential risks**


yuandrew:
We don't need to manually pull in updates from the `api` repo, we have a GH action that runs when updates are merged to main that will automatically update `api-go` https://github.com/temporalio/api/actions/workflows/trigger-api-go-update.yml

PR #255, opened 2026-04-27
Add Standalone Callbacks (update proxygenerator, proto/api SHA)

chrsmith:
> This is part of the standalone callbacks / manual completions workstream. And should be merged once the API and server-side bits are ready for merging as well.
>
> As-is, the SHA referenced in `proto/api` points to an unmerged PR https://github.com/temporalio/api/pull/774.

**What changed?**

Updates the Golang API package to support standalone callbacks.

**Why?**

- Bump the `proto/api` is required to pickup the protos while the underlying protos haven't made it to `main` yet.

- Update `cmd/proxygenerator` to support a new protobuf message using a `oneof` field. I don't know exactly why this is required yet. But the generated code fails to compile without it.

- Updates the `serviceerror` package to expose a well-defined service error to be consistent with other similar errors.

**How did you test it?**

Locally, and in the server-side and SDK changes.

**Potential risks**

None I can think of.


serviceerror/callback_execution_already_started.go
```diff
@@ -0,0 +1,70 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"go.temporal.io/api/errordetails/v1"
+)
+
+// CallbackExecutionAlreadyStarted represents the error arising when trying to start a callback already in progress.
+type CallbackExecutionAlreadyStarted struct {
+	Message        string
+	StartRequestID string
```
stephanos, line 15: It appears the other variants of this use `StartRequestId` (lowercase `d`)

serviceerror/nexus_operation_not_started.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"go.temporal.io/api/errordetails/v1"
+)
+
+type (
+	// NexusOperationNotStarted represents a nexus operation execution already started error.
```
stephanos, line 13: Isn't this for when it's _not_ started?

serviceerror/nexus_operation_not_started.go
```diff
@@ -0,0 +1,62 @@
...
+func (e *NexusOperationNotStarted) Error() string {
+	return e.Message
+}
+
+func (e *NexusOperationNotStarted) Status() *status.Status {
+	if e.st != nil {
+		return e.st
+	}
+
+	st := status.New(codes.AlreadyExists, e.Message)
```
stephanos, line 47: I thought it's `codes.FailedPrecondition`?

chrsmith:
This PR is pretty much complete, however we opted to go with a different approach for implementing standalone completions. (And will look into using SAA for this functionality.) This code might be useful down the road, but this particular PR can be closed out.

PR #290, opened 2026-06-23
Recursively visit payloads inside ScheduleNexusOperationCommandAttributes when using system nexus

dplyukhin:
## What

System Nexus messages shouldn't go through the codec because they're meant to be consumed by the Temporal service itself. But payloads *inside* the message should go through the codec.

## How

Modify the payload visitor to treat `ScheduleNexusOperationCommandAttributes` commands in a special way if the endpoint is `__temporal_system`. Specifically, we deserialize them using proto type information in the metadata, and recursively visiting the payloads inside. This is similar to what we already do for `anypb.Any` messages with the `defaultWellKnownAnyVisitor`.

⚠️ This PR only works on the *inputs* for system nexus operations. The output of a system nexus operation doesn't have proto type metadata attached, so we'll have to do something else there.

## Tests

- [x] `make proxy && (cd cmd/proxygenerator && go run ./ -verifyOnly)` (generated code is current)
- [x] `go test -tags protolegacy ./...`
- [x] Add tests to `proxy/system_nexus_test.go` that exercise the happy and sad paths for visiting system nexus payloads


Makefile
```diff
@@ -97,10 +85,37 @@ grpc-mock:
...
+.PHONY: nexus-system-registry
+nexus-system-registry: gen-proto-desc
+	printf $(COLOR) "Generate Nexus system operation registry..."
+	rm -rf $(NEXUS_SYSTEM_DIR)/gen
+	$(NEX_GEN) generate \
+		--lang go \
+		--input $(NEXUS_SYSTEM_WIT) \
+		--input $(NEXUS_SYSTEM_WIT_DEPS) \
+		--descriptors $(PROTO_OUT)/descriptor_set.pb \
+		--output $(NEXUS_SYSTEM_DIR)/gen
```
tconley1428: We should probably add an opt in or out of API generation.

proxy/system_nexus.go
```diff
@@ -0,0 +1,129 @@
...
+	"go.temporal.io/api/command/v1"
+	"go.temporal.io/api/nexussystem"
+	"google.golang.org/protobuf/proto"
+)
+
+// systemNexusServiceName is the Nexus service used for system Nexus envelopes
+// routed to the workflow service. Together with nexussystem.Endpoint it
+// identifies a ScheduleNexusOperationCommandAttributes whose Input is a system
+// Nexus envelope.
+const systemNexusServiceName = "temporal.api.workflowservice.v1.WorkflowService"
```
tconley1428: This needs to be removed.

proxy/system_nexus.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+// isSystemNexusEnvelope reports whether the given schedule-nexus-operation
+// command targets the system Nexus endpoint and service, and therefore carries
+// a proto-message envelope in its Input rather than an opaque user payload.
+func isSystemNexusEnvelope(attrs *command.ScheduleNexusOperationCommandAttributes) bool {
+	if attrs == nil {
+		return false
+	}
+	return attrs.GetEndpoint() == nexussystem.Endpoint &&
```
tconley1428: Let's standardize on just checking the endpoint and using the encoded type metadata.

tconley1428 (approved):
Let's mention that it only handles system nexus _inputs_ currently. We can't fix that quite yet but it would be good to remember.

proxy/system_nexus.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	// Sub-state shares the semaphore but has its own WaitGroup so we can wait
+	// for goroutines writing into child's fields before re-marshaling.
+	var localConcState *payloadConcurrencyState
+	if concState != nil {
+		localConcState = &payloadConcurrencyState{sem: concState.sem}
+	}
+
+	if err := visitPayloads(ctx, options, msg, localConcState, msg); err != nil {
```
yuandrew: I think this nested concurrent path should follow the top-level `VisitPayloads` wait-before-return shape. At the top level we do:

`proxy/interceptor.go:110`
  ```go
  err := visitPayloads(&visitCtx, &options, nil, c, msg)
  c.wg.Wait()
  if err != nil {
      return err
  }
```

but here the function returns before waiting if recursive traversal returns an error.

Since localConcState can already have spawned visitor goroutines using the shared semaphore, an early traversal error or cancellation can let those goroutines continue after VisitPayloads returns.

There is a similar nested waitgroup pattern in defaultWellKnownAnyVisitor around proxy/interceptor.go:281, but it has the same early-return shape. I would use the top-level pattern here instead: store the traversal error, always wait on localConcState.wg, then return the traversal error or localConcState.firstErr.

dplyukhin: Nice catch! Fixed.

PR #293, opened 2026-07-07
Add WorkflowTaskBufferLost serviceerror

simvlad:
Adds the `WorkflowTaskBufferLost` serviceerror

The proto/errordetails for paginated workflow task completion already landed, but the hand-written serviceerror wrapper (the actual returnable Go error) was missing. The server returns this transient error when it loses the
buffered intermediate pages of a paginated completion, signaling the client to resend all pages from page 0 with the same task token.

`go build`

**Potential risks**
Additive change


serviceerror/workflow_task_buffer_lost.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"go.temporal.io/api/errordetails/v1"
+)
+
+type (
+	// WorkflowTaskBufferLost represents a lost buffer during processing of a pagination requestt.
```
jmaeagle99: nit: The name of the struct is more generic than what the comment says how it is intended to be used. The current name suggests that it could be used for any buffer that is associated with a workflow task. But this seems to be consistent with the name `WorkflowTaskCompletionBufferLostFailure`.

jmaeagle99:
```suggestion
	// WorkflowTaskBufferLost represents a lost buffer during processing of a pagination request.
```

jmaeagle99: Maybe rename to `WorkflowTaskCompletionBufferLost` to be consistent with the other error types that just drop the `Failure` suffix. And update the functions that are similarly named.

jmaeagle99: Should the comments state that clients should restart pagination from page 0?

jmaeagle99 (approved):

PR #298, opened 2026-07-20
Add link support for Activity and Workflow

Quinn-With-Two-Ns:
**What changed?**

Add link support for Activity and Workflow

**Why?**

So the SDK , CLI, and Server can all share common link handling code

**How did you test it?**

**Potential risks**

### DIFF
--- a/temporalnexus/link_converter.go
+++ b/temporalnexus/link_converter.go
@@ -25,11 +25,13 @@ const (
 	urlPathWorkflowEventTemplate  = "/namespaces/%s/workflows/%s/%s/history"
 	urlPathNexusOperationTemplate = "/namespaces/%s/nexus-operations/%s/%s/details"
 	urlPathActivityTemplate       = "/namespaces/%s/activities/%s/%s/details"
+	urlPathWorkflowTemplate       = "/namespaces/%s/workflows/%s/%s"
 
 	linkWorkflowEventReferenceTypeKey = "referenceType"
 	linkEventIDKey                    = "eventID"
 	linkEventTypeKey                  = "eventType"
 	linkRequestIDKey                  = "requestID"
+	linkReasonKey                     = "reason"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


temporalnexus/link_converter.go
```diff
@@ -56,11 +58,19 @@ var (
 		rePatternActivityID,
 		rePatternRunID,
 	))
+	urlPathWorkflowRE = regexp.MustCompile(fmt.Sprintf(
+		`^/namespaces/%s/workflows/%s/%s$`,
```
mavemuri, line 65: nit: maybe re-use the urlPathWorkflowTemplate here so its clearer/doesnt drift

temporalnexus/link_converter.go
```diff
@@ -298,6 +308,83 @@ func ConvertNexusLinkToLinkActivity(link nexus.Link) (*commonpb.Link_Activity, e
...
+		Type: string(w.ProtoReflect().Descriptor().FullName()),
+	}
+}
+
+// ConvertNexusLinkToLinkWorkflow converts a Nexus Link to Link_Workflow.
+//
+// NOTE: Experimental
+func ConvertNexusLinkToLinkWorkflow(link nexus.Link) (*commonpb.Link_Workflow, error) {
+	w := &commonpb.Link_Workflow{}
+	if link.Type != string(w.ProtoReflect().Descriptor().FullName()) {
```
mavemuri: nit: maybe use the workflowLinkType above- shouldnt ever happen for anyone using the common function though(also avoids minor w alloc but thats probably negligible)

temporalnexus/link_converter.go
```diff
@@ -298,6 +308,83 @@ func ConvertNexusLinkToLinkActivity(link nexus.Link) (*commonpb.Link_Activity, e
...
+		Path:   fmt.Sprintf(urlPathWorkflowTemplate, w.GetNamespace(), w.GetWorkflowId(), w.GetRunId()),
+		RawPath: fmt.Sprintf(
+			urlPathWorkflowTemplate,
+			url.PathEscape(w.GetNamespace()),
+			url.PathEscape(w.GetWorkflowId()),
+			url.PathEscape(w.GetRunId()),
+		),
+	}
+
+	if w.GetReason() != "" {
```
mavemuri, line 329: question: i dont know what the "reason" indicates - can you please point to docs/other? Maybe this is obvious but why do we avoid adding a fallback/default reason given that we know its a Nexus link?

mavemuri (approved):

temporalnexus/link_converter.go
Quinn-With-Two-Ns: Done, updates the other converter functions to so they are all consistent

temporalnexus/link_converter.go
Quinn-With-Two-Ns: "reason" is just the reason the link was created, there is no logical default reason

PR #305, opened 2026-07-29
Update Makefile to copy WIT files from temporal-api

dplyukhin:
**What changed?**
Updated Makefile to copy WIT service definitions from temporal-api

**Why?**
So the Go SDK can use the files to generate system nexus APIs like `SignalWithStartWorkflow`.

**How did you test it?**
Tested locally

**Potential risks**


Makefile
```diff
@@ -42,9 +42,10 @@ update-proto-submodule:
 
 NEXUS_SCHEMA_ROOT := $(PROTO_ROOT)/nexus
 NEXUS_PROTO_OUT := workflowservice/v1/workflowservicenexus
+NEXUS_WIT_OUT := workflowservice/v1/systemnexus
```
dplyukhin: Does anyone have opinions on where we should put the WIT files?

Makefile
```diff
@@ -55,6 +56,13 @@ nexus-gen:
 		--out-file $(CURDIR)/$(NEXUS_PROTO_OUT)/service_nexus.pb.go \
 		temporal-proto-models-nexusrpc.yaml
 
+.PHONY: system-nexus
+system-nexus:
+	printf $(COLOR) "Update system-nexus WIT files..."
+	rm -rf $(NEXUS_WIT_OUT)
+	mkdir -p $(NEXUS_WIT_OUT)
+	cp -R $(NEXUS_SCHEMA_ROOT)/workflow-service.wit $(NEXUS_SCHEMA_ROOT)/deps $(NEXUS_WIT_OUT)/
```
dplyukhin: @tconley1428 Based on your plans for future WIT stuff, does this make sense for how to lay things out in the api-go directory? i.e., we copy `XXX-service.wit` into the `XXXservice/v1/systemnexus` directory, along with the entirety of `deps/`?

tconley1428: I don't really have any opinion on location here.

yuandrew (approved):
Looks reasonable to me, @Quinn-With-Two-Ns, not sure if you wanna also take a peak

Makefile
Quinn-With-Two-Ns: just speculating, if we did add a `systemnexus` to the gRPC workflow service would they end up conflicting?

